PR #3, opened 2020-10-27
Adds code

JoshVanL:

/assign @james-w 
/assign @jakexks 
/assign @meyskens 
/assign @wallrj


Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM alpine:3.12
```
meyskens, line 1: We use distroless everywhere, is there reasoning for using alpine here? From a conformance standpoint would be better to use one base image?

README.md
```diff
@@ -1 +1,53 @@
-# istio-csr
\ No newline at end of file
+# istio-csr
+
+cert-manager-istio-agent is an agent which allows for [istio](istio.io) workload
```
meyskens, line 3:
```suggestion
cert-manager-istio-agent is an agent which allows for [istio](https://istio.io) workload
```

cmd/app/options/options.go
```diff
@@ -0,0 +1,186 @@
...
+		Name:  o.issuerName,
+		Kind:  o.issuerKind,
+		Group: o.issuerGroup,
+	}
+
+	return nil
+}
+
+func (o *Options) addFlags(cmd *cobra.Command) {
+	var nfs cliflag.NamedFlagSets
```
meyskens, line 117: This is a way of flags i haven't seen before? Is this something specific?

deploy/charts/istio-csr/templates/certificate.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: istiod
+  namespace: istio-system
+spec:
+  isCA: false
```
meyskens, line 7: can be left out imo

deploy/charts/istio-csr/templates/deployment.yaml
```diff
@@ -0,0 +1,39 @@
...
+kind: Deployment
+metadata:
+  name: {{ include "cert-manager-istio-agent.name" . }}
+  labels:
+{{ include "cert-manager-istio-agent.labels" . | indent 4 }}
+spec:
+  replicas: {{ .Values.replicaCount }}
+  selector:
+    matchLabels:
+      app: {{ include "cert-manager-istio-agent.name" . }}
```
meyskens, line 11: Can we add more labels to match on?

pkg/controller/namespace.go
```diff
@@ -0,0 +1,271 @@
...
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/leaderelection"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
+	"k8s.io/client-go/util/workqueue"
+
+	"github.com/cert-manager/istio-csr/cmd/app/options"
+)
+
+const (
+	resyncPeriod        = time.Second * 60
```
meyskens, line 25: recent cert-manager discussions linking to controller-runtime work indicates that this can be waaay longer ;)

pkg/server/auth.go
```diff
@@ -0,0 +1,82 @@
...
+	identities := strings.Join(caller.Identities, ",")
+
+	csr, err := pkiutil.ParsePemEncodedCSR(csrPEM)
+	if err != nil {
+		s.log.Errorf("failed to decode CSR from %s: %s", identities, err)
+		return identities, false
+	}
+
+	// if the csr contains any other options set, error
+	if len(csr.DNSNames) > 0 || len(csr.IPAddresses) > 0 ||
```
meyskens, line 38: Is this general checking or istio specific? it doesn't check for URI SANs

meyskens, line 38: I retract my comment, it was caused by the lack of coffee

pkg/server/server.go
```diff
@@ -0,0 +1,159 @@
...
+	identities, ok := s.authRequest(ctx, []byte(icr.Csr))
+	if !ok {
+		return nil, status.Error(codes.Unauthenticated, "request authenticate failure")
+	}
+
+	// Build cert-manager CertificateRequest based on the configured issuer
+	cr := &cmapi.CertificateRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			// Random non-conflicted name
+			GenerateName: "istio-",
```
meyskens, line 92: do we have some name from the request? Thinking about a user debugging seeing 100 CRs called istio

pkg/tls/tls.go
```diff
@@ -0,0 +1,281 @@
...
+		return nil, errors.New("provider not configured, TLS config not ready")
+	}
+
+	return &tls.Config{
+		GetConfigForClient: p.getConfigForClient,
+		ClientAuth:         tls.RequireAndVerifyClientCert,
+	}, nil
+}
+
+// hetConfigForClient will return a TLS config based upon the current signed
```
meyskens, line 145:
```suggestion
// getConfigForClient will return a TLS config based upon the current signed
```

pkg/tls/tls.go
```diff
@@ -0,0 +1,281 @@
...
+				"istio.cert-manager.io/identities": "cert-manager-istio-agent",
+			},
+		},
+		Spec: cmapi.CertificateRequestSpec{
+			Duration: &metav1.Duration{
+				Duration: p.servingCertificateTTL,
+			},
+			IsCA:      false,
+			Request:   csr,
+			Usages:    []cmapi.KeyUsage{cmapi.UsageServerAuth},
```
meyskens, line 194: I do wonder here, is this enough? this disables all (classic) KeyUsages?

Dockerfile
JoshVanL: I agree, I'll fix that :)

cmd/app/options/options.go
JoshVanL: It's basically used for pretty printing flags into names groupings;

```bash
Usage:
  cert-manager-istio-agent [flags]

App flags:

  -v, --log-level string   Log level (debug, info, warn, error, fatal, panic). (default "info")

TLS flags:

      --root-ca-cert string                File location of a PEM encoded Root CA certificate to be used as root of trust for TLS. If empty, the CA returned from the cert-manager issuer will be used.
  -a, --serving-address string             Address to serve certificates gRPC service. (default "0.0.0.0:443")
  -t, --serving-certificate-ttl duration   TTL duration of serving certificates. Will be renewed after 2/3 of the duration. (default 24h0m0s)

Cert-manager flags:

  -c, --certificate-namespace string    Namespace to request certificates. (default "istio-system")
  -g, --issuer-group string             Group of the issuer to sign istio workload certificates. (default "cert-manager.io")
  -k, --issuer-kind string              Kind of the issuer to sign istio workload certificates. (default "Issuer")
  -u, --issuer-name string              Name of the issuer to sign istio workload certificates. (default "istio-ca")
  -d, --preserve-certificate-requests   If enabled, will preserve created CertificateRequests, rather than deleting when they are ready.

Kubernetes flags:

      --as string                      Username to impersonate for the operation
      --as-group stringArray           Group to impersonate for the operation, this flag can be repeated to specify multiple groups.
....................
```

deploy/charts/istio-csr/templates/deployment.yaml
JoshVanL: What else would you like to match on?

pkg/controller/namespace.go
JoshVanL: How come, are their concerns about syncing too much?

Could you expand a bit more?

pkg/server/auth.go
JoshVanL: :coffee:

pkg/server/server.go
JoshVanL: Yeah, we have the requester spiffe identity, which is a part of the annotation (mostly for auditing).

These id's can get too long for names.

pkg/tls/tls.go
JoshVanL: This is enough for my testing. This is just for serving the cert service for istio proxies (envoy). This passes e2e tests with 1.7

deploy/charts/istio-csr/templates/deployment.yaml
meyskens: need to look at what we do in cert-manager (core) i see manuy re-use the labels helper

pkg/controller/namespace.go
meyskens: long story... becomes problematic when talking 10K resources

meyskens:
Reviewed to the best of my ability :) looks good in general!

Makefile
```diff
@@ -0,0 +1,46 @@
+BINDIR ?= $(CURDIR)/bin
+ARCH   ?= amd64
+ISTIO_VERSION ?= 1.7.3
+DEMO_MANIFEST_URL ?= https://raw.githubusercontent.com/istio/istio/master/samples/bookinfo/platform/kube/bookinfo.yaml
+
+help:  ## display this help
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n\nTargets:\n"} /^[a-zA-Z0-9_-]+:.*?##/ { printf "  \033[36m%-20s\033[0m %s\n", $$1, $$2 }' $(MAKEFILE_LIST)
+
+.PHONY: help build docker all clean
```
james-w, line 9:
```suggestion
.PHONY: help test build verify image clean all demo docker e2e depend
```

README.md
```diff
@@ -1 +1,53 @@
...
+
+Currently supports istio v1.7
+
+---
+
+## Installation
+
+Firstly, [cert-manager must be
+installed](https://cert-manager.io/docs/installation/) in your cluster. An
+issuer must be configured, which will be used to sign your certificate
```
james-w, line 17: Are there any important constraints on the issuer? e.g. does it have to be a clusterissuer?

deploy/charts/istio-csr/templates/certificate.yaml
```diff
@@ -0,0 +1,17 @@
...
+kind: Certificate
+metadata:
+  name: istiod
+  namespace: istio-system
+spec:
+  duration: 24h
+  renewBefore: 12h
+  secretName: istiod-tls
+  dnsNames:
+  - istiod.istio-system.svc
```
james-w, line 11: Would this ever change? Probably not important for now, but is it required to use `istio-system` for your istio installation?

pkg/controller/namespace.go
```diff
@@ -0,0 +1,271 @@
...
+
+			OnStoppedLeading: func() {
+				c.log.Info("lost leader election")
+				cancel()
+			},
+		},
+	})
+}
+
+func (c *CARoot) runController(ctx context.Context) error {
```
james-w, line 100: You didn't fancy using controller-runtime?

pkg/server/auth_test.go
```diff
@@ -0,0 +1,239 @@
...
+func genCSR(t *testing.T, mods ...gen.CSRModifier) []byte {
+	csr, err := gen.CSR(mods...)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	return csr
+}
+
+func TestAuthRequest(t *testing.T) {
```
james-w, line 110: Great to see tests of this.

pkg/server/server.go
```diff
@@ -0,0 +1,159 @@
...
+	cr, err := s.client.Create(ctx, cr, metav1.CreateOptions{})
+	if err != nil {
+		s.log.Errorf("failed to create CertificateRequest for %q: %s",
+			identities, err)
+		return nil, status.Error(codes.Internal, "failed to sign certificate request")
+	}
+
+	log := util.LogWithCertificateRequest(s.log, cr)
+
+	cr, err = util.WaitForCertificateRequestReady(ctx, log, s.client, cr.Name, time.Second*30)
```
james-w, line 122: Should this be configurable? Is there some problem if certificaterequests take longer to be signed? I guess that istio will timeout waiting for us at some point anyway?

pkg/server/server.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		return nil, status.Error(codes.DeadlineExceeded, "timeout exceeded waiting for certificate request to be signed")
```
james-w, line 124: I guess we may want some cleanup process to deal with leaked CRs at some point?

test/e2e/suite/request/request.go
```diff
@@ -0,0 +1,260 @@
...
+		saToken = string(saTokenBytes)
+	})
+
+	JustAfterEach(func() {
+		By("removing test namespace with service account token")
+		err := f.KubeClientSet.CoreV1().Namespaces().Delete(context.TODO(), namespace, metav1.DeleteOptions{})
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("should reject a request with a bad service account token", func() {
```
james-w, line 96: Thanks for writing tests :)

README.md
JoshVanL: No, you can have whatever you want:

```
  -g, --issuer-group string             Group of the issuer to sign istio workload certificates. (default "cert-manager.io")
  -k, --issuer-kind string              Kind of the issuer to sign istio workload certificates. (default "Issuer")
  -u, --issuer-name string              Name of the issuer to sign istio workload certificates. (default "istio-ca")
```

deploy/charts/istio-csr/templates/certificate.yaml
JoshVanL: As far as I know it's required, and basically hardcoded

README.md
james-w: If it's an Issuer presumably it has to be in the namespace where you are running the agent? That's probably obvious?

pkg/controller/namespace.go
JoshVanL: This is mostly a copy of https://github.com/istio/istio/blob/master/security/pkg/k8s/controller/casecret.go

with a couple things changed

pkg/server/server.go
JoshVanL: Yes, I think we should extend this though anyhow. As you say, we will start to balloon CRs if the issuer is broken. I will increase the time limit and delete if errored?

README.md
JoshVanL: No, in the default installation, the agent runs in cert-manager, CRs and Issuer are in istio-system.

You can configure the target namespace with;

```bash
  -c, --certificate-namespace string    Namespace to request certificates. (default "istio-system")
```

I would have thought the Issuer being in the istio-system being obvious but perhaps not..

pkg/server/server.go
james-w: That seems like a fine start. There's a flag to retain the certificaterequests for debugging.

README.md
james-w: A good rule of thumb is that nothing is as obvious as you think :)

JoshVanL: A rule to live by.. I'll update the README.md

hack/istio-config.yaml
```diff
@@ -0,0 +1,41 @@
...
+          # Disable istiod CA Sever functionality
+        - name: ENABLE_CA_SERVER
+          value: "false"
+        overlays:
+        - kind: Deployment
+          name: istiod
+          patches:
+
+            # Mount istiod serving and webhook certificate from Secret mount
+          - path: spec.template.spec.containers.[name:discovery].args[8]
```
shankgan, line 24: Is there any way to append to the list of arguments already provided as opposed to specifying the n'th argument?

JoshVanL: No, this is a known issue https://github.com/istio/istio/issues/21863

Would like to know a better work around because I'm not a fan of this either

pkg/server/server.go
```diff
@@ -0,0 +1,159 @@
...
+				// Label identities to resource for auditing
+				IdentitiesAnnotationKey: identities,
+			},
+		},
+		Spec: cmapi.CertificateRequestSpec{
+			Duration: &metav1.Duration{
+				// Add during which was requested from the client.
+				// TODO: We should have a configurable maximum that the duration can
+				// be. Take smaller of the two.
+				Duration: time.Duration(icr.ValidityDuration) * time.Second,
```
JoshVanL, line 103: Is this appropriate?

shankgan: Given that the agent could function as a local CA, it makes sense to have a flag that configures the maximum possible certificate duration

PR #22, opened 2021-01-27
Add copywrite header

JoshVanL:


JoshVanL:
Will looks into why this task is failing.

/retest

PR #40, opened 2021-03-17
Adds wait to namespace controller propagation test

JoshVanL:

Adds a wait to the e2e namespace controller test for waiting to root CA propagation. This prevents flakes from happening in e2e tests.

/assign @SgtCoDFish @jakexks @wallrj @irbekrm


SgtCoDFish:
This looks good!

A minute feels like quite a long timeout for these condition to be true (although I've not measured it)? I don't think that timeout would be a blocker to merging though.

JoshVanL:
@SgtCoDFish Yeah, I have reduced to 30 secs

PR #46, opened 2021-04-13
Fix istio config for v1.9.1

### LINKED ISSUE #45: Wrong example
https://github.com/cert-manager/istio-csr/blob/9a223f9222924a5d65b90742391f293c900c7ce2/hack/istio-config-1.9.1.yaml#L52-L55

Duplicate _name_.

m1pl:
fixes #45

### DIFF
--- a/hack/istio-config-1.9.1.yaml
+++ b/hack/istio-config-1.9.1.yaml
@@ -50,6 +50,5 @@ spec:
             value:
               name: ca-root-cert
               configMap:
-                name: istiod-tls
                 defaultMode: 420
                 name: istio-ca-root-cert
--- a/hack/istio-config-1.7.6.yaml
+++ b/hack/istio-config-1.7.6.yaml
@@ -50,6 +50,5 @@ spec:
             value:
               name: ca-root-cert
               configMap:
-                secretName: istiod-tls
                 defaultMode: 420
                 name: istio-ca-root-cert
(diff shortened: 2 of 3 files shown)


m1pl:
CONTRIBUTING.md is missing...

hack/istio-config-1.9.1.yaml
```diff
@@ -50,6 +50,6 @@ spec:
             value:
               name: ca-root-cert
               configMap:
-                name: istiod-tls
+                secretName: istiod-tls
```
JoshVanL, line 53: Thanks! I think this line needs to be removed completely?

JoshVanL (changes requested):
Thanks for this!

I will add a `CONTRIBUTING.md`.

/approve
/ok-to-test

hack/istio-config-1.9.1.yaml
m1pl: @JoshVanL I copied it from the old version: https://github.com/cert-manager/istio-csr/blob/dacb33e62cd33ff6f5787d371de35666d5d29511/hack/istio-config-1.8.2.yaml#L53

JoshVanL: Yep, I think it is also wrong there as well.

m1pl: So you want me to remove all three lines, or just the one for 1.9.1?

m1pl:
@JoshVanL I removed the line from all 3 files.

JoshVanL:
Thanks @m1pl!

/lgtm

PR #51, opened 2021-05-07
Allow the TrustDomain to be set.

janzantinge:
This PR is meant to add support for custom trust domains to Istio CSR. It sets the domain in the SPIFFE library allowing for certificates with a the trust domain to be signed.

The tests were updated to specify a custom trust domain (`foo.bar`). LMK if there's a better way to test a change like this.

The Helm chart was updated as well, a new `value` was added: `trustDomain`.

### DIFF
--- a/cmd/app/options/options.go
+++ b/cmd/app/options/options.go
@@ -75,7 +75,8 @@ type TLSOptions struct {
 	ServingAddress             string
 	ServingCertificateDuration time.Duration
 
-	ClusterID string
+	ClusterID   string
+	TrustDomain string
 }
 
 type KubeOptions struct {
@@ -106,6 +107,11 @@ func (o *Options) Complete() error {
 	log := klogr.New()
 	flag.Set("v", o.logLevel)
 	o.Logr = log
+
+	// Set the trust domain before the Auther and tls Provider are created to
+	// ensure the trust domain is set correctly before being used to
+	// authenticate requests
+	spiffe.SetTrustDomain(o.TLSOptions.TrustDomain)
 
 	var err error
 	o.RestConfig, err = o.kubeConfigFlags.ToRESTConfig()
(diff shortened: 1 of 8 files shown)


cmd/app/options/options.go
```diff
@@ -67,6 +67,8 @@ type CertManagerOptions struct {
 	Namespace   string
 	PreserveCRs bool
 	IssuerRef   cmmeta.ObjectReference
+
+	TrustDomain string
```
JoshVanL, line 71: I think this is better suited to live under `TLSOptions.TrustDomain`.

cmd/app/options/options.go
```diff
@@ -225,4 +232,8 @@ func (c *CertManagerOptions) addFlags(fs *pflag.FlagSet) {
 	fs.StringVarP(&c.Namespace,
 		"certificate-namespace", "c", "istio-system",
 		"Namespace to request certificates.")
+
+	fs.StringVar(&c.TrustDomain,
+		"trust-domain", "cluster.local",
+		"The cluster's trust domain.")
```
JoshVanL, line 238:
```suggestion
		"The istio cluster's trust domain.")
```

deploy/charts/istio-csr/templates/deployment.yaml
```diff
@@ -43,6 +43,7 @@ spec:
           - "--issuer-group={{.Values.certificate.group}}"
           - "--issuer-kind={{.Values.certificate.kind}}"
           - "--issuer-name={{.Values.certificate.name}}"
+          - "--trust-domain={{.Values.trustDomain}}"
```
JoshVanL, line 46:
```suggestion
          - "--trust-domain={{.Values.agent.trustDomain}}"
```

JoshVanL (changes requested):
Thanks @janzantinge! Just a couple comments from me.

/unassign
/assign @janzantinge 
/approve
/ok-to-test

hack/istio-config-1.7.6.yaml
```diff
@@ -5,8 +5,16 @@ metadata:
...
+  meshConfig:
+    # Change the following lines to configure the trust domain of the Istio cluster.
+    trustDomain: foo.bar
+    defaultConfig:
+      proxyMetadata:
+        TRUST_DOMAIN: foo.bar
   values:
     global:
+      # Change the following line to configure the trust domain of the Istio cluster.
+      trustDomain: foo.bar
```
janzantinge, line 17: This took me a bit of time to figure out, updating all of these fields is required before Istio v1.8.0.

JoshVanL:
Thanks @janzantinge! This is a much needed addition.

/lgtm

Issue #36, opened 2021-03-10
readiness probe failed


adnankobir:
After running istio-csr for some time, some pods fail readiness probe:

```
❯ kubectl get po -n cert-manager
NAME                                                    READY   STATUS    RESTARTS   AGE
cert-manager-istio-csr-79ffc5bfd-dm7mv                  1/1     Running   0          7d17h
cert-manager-istio-csr-79ffc5bfd-mv7wj                  0/1     Running   0          20d
cert-manager-istio-csr-79ffc5bfd-t855r                  1/1     Running   0          20d
```

Describing the pod:
```
❯ kubectl describe po cert-manager-istio-csr-79ffc5bfd-mv7wj -n cert-manager
Name:         cert-manager-istio-csr-79ffc5bfd-mv7wj
Namespace:    cert-manager
Priority:     0
Node:         ip-10-135-223-54.ec2.internal/10.135.223.54
Start Time:   Wed, 17 Feb 2021 12:28:14 -0500
Labels:       app=cert-manager-istio-csr
              pod-template-hash=79ffc5bfd
Annotations:  kubernetes.io/psp: eks.privileged
Status:       Running
IP:           10.135.215.165
IPs:
  IP:           10.135.215.165
Controlled By:  ReplicaSet/cert-manager-istio-csr-79ffc5bfd
Containers:
  cert-manager-istio-csr:
    Container ID:  docker://2616a2dd356d0b6502b32da6589cae521744e3b22ecc730a8bcde738b50bbaab
    Image:         quay.io/jetstack/cert-manager-istio-csr:v0.1.0
    Image ID:      docker-pullable://quay.io/jetstack/cert-manager-istio-csr@sha256:f9d473fa10520d0a255a4b60350a9f9057834da762129f9e5ecb9681955b1fd0
    Port:          6443/TCP
    Host Port:     0/TCP
    Command:
      cert-manager-istio-csr
    Args:
      --log-level=1
      --readiness-probe-port=6060
      --readiness-probe-path=/readyz
      --serving-address=0.0.0.0:6443
      --serving-certificate-duration=24h
      --root-ca-configmap-name=istio-ca-root-cert
      --certificate-namespace=istio-system
      --issuer-group=cert-manager.io
      --issuer-kind=ClusterIssuer
      --issuer-name=vault-issuer
      --max-client-certificate-duration=24h
      --preserve-certificate-requests=false
    State:          Running
[...]

PR #62, opened 2021-06-10
Exit error if we lose leader election, fix election ID

JoshVanL:

This PR fixes a bug where is the leader election is lost because of a transient network error or similar, the istio-csr becomes unready. To fix this issue, istio-csr will exit error, which will cause either: 1. another istio-csr to assume the leader, or 2. istio-csr restarts and assumes the leader again.

Exiting the program is the [canonical way](https://github.com/kubernetes-sigs/controller-runtime/blob/fbf50b04fe17b0f091430d22557245d2aedabfd4/pkg/manager/manager.go#L83) for handling losing the leader election. 

Adding a specific event broadcaster is used to prevent a permissions error where istio-csr would create an event in the default namespace when losing/failing to renew a lease. By passing a custom recorder, we force the event to be created in the istio-system namespace which istio-csr has permissions for.

/assign @jakexks 
fixes #20 #36

### DIFF
--- a/cmd/app/app.go
+++ b/cmd/app/app.go
@@ -19,6 +19,7 @@ package app
 import (
 	"context"
 	"fmt"
+	"os"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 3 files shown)


JoshVanL:
```
  Warning  Failed     8m20s                 kubelet            Failed to pull image "gcr.io/istio-release/proxyv2:1.9.1": rpc error: code = Unknown desc = failed to pull and unpack image "gcr.io/istio-release/proxyv2:1.9.1": failed to resolve reference "gcr.io/istio-release/proxyv2:1.9.1": failed to do request: Head "https://gcr.io/v2/istio-release/proxyv2/manifests/1.9.1": dial tcp: lookup gcr.io on 172.18.0.1:53: read udp 172.18.0.2:44030->172.18.0.1:53: i/o timeout
```

/retest

PR #67, opened 2021-06-21
Update Deps

JoshVanL:
This PR updates the istio and cert-manager dependencies to v1.10 and v1.4 respectively.

This introduces a new "mesh.Holder" interface which is used to initialize the kube auther with the correct trust domain. 

The internal CA client has been changed according to the new client interface.

The request e2e test has been refactored to reflect the internal client changes.

/assign @jakexks


test/e2e/suite/request/request.go
```diff
@@ -258,19 +271,19 @@ var _ = framework.CasesDescribe("Request Authentication", func() {
 		}
 
 		if createdCR == nil {
-			Expect("did not find created CertificateRequest for identity").NotTo(HaveOccurred())
+			Fail(fmt.Sprintf("did not find created CertificateRequest for identity %q", id))
```
irbekrm, line 274: Nice! The `Fail` function is new in ginkgo I guess.

test/e2e/suite/internal/client/client.go
```diff
@@ -70,35 +73,22 @@ func NewCertManagerClient(endpoint string, tls bool, rootCert []byte, clusterID
...
 		Csr:              string(csrPEM),
 		ValidityDuration: certValidTTLInSec,
 	}
-
-	if token != "" {
-		// add Bearer prefix, which is required by cert-manager istio agent.
-		token = bearerTokenPrefix + token
-		ctx = metadata.NewOutgoingContext(ctx, metadata.Pairs("Authorization", token, "ClusterID", c.clusterID))
-	} else {
-		// This may use the per call credentials, if enabled.
```
irbekrm, line 85: Question: is this an option that was never implemented in istio-csr anyway? (I see all the e2e tests always used a token).

irbekrm (commented):
Looks good to me- I've read through the changes here and in istio.
I've only added a few of questions to make sure I've not missed/misunderstood something- happy to lgtm otherwise.

cmd/app/options/options.go
```diff
@@ -115,11 +114,6 @@ func (o *Options) Complete() error {
 		return fmt.Errorf("the list of DNS names to add to the serving certificate is empty")
 	}
 
-	// Set the trust domain before the Auther and tls Provider are created to
-	// ensure the trust domain is set correctly before being used to
-	// authenticate requests
-	spiffe.SetTrustDomain(o.TLSOptions.TrustDomain)
```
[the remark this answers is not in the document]
JoshVanL: Interesting. I want to say that we don't care about this for now. If this becomes an issue, we can revisit.

There is already a workaround by the user doing this string replace as part of the TLSOptions config.

test/e2e/suite/internal/client/client.go
JoshVanL: Correct. This file was adapted from [istio's own caClient](https://github.com/istio/istio/blob/c3d92d81db5dd5b2cba7211db25f0ab8b51a6d3b/security/pkg/nodeagent/caclient/providers/citadel/client.go#L41). Removing unused bits is always good :factory_worker:

irbekrm:
Thanks for answering my questions!

Looks good as far as I can tell 😄 

/lgtm

PR #68, opened 2021-06-21
Remove test boilerplate files

JoshVanL:
Remove boilerplate test files which would cause `go test ./...` to fail.

/assign @jakexks


irbekrm:
Makes sense, I see that we have removed the `hack/boilerplate/test` in cert-manager too.

/lgtm

PR #70, opened 2021-06-21
Change istio config to use "append last" in patches

JoshVanL:

Updates the istio configs so that changes to args/volumes/mounts use an append strategy. This helps to prevent overwriting current values.

/assign @jakexks


irbekrm:
Looks good to me- I haven't tried it, but looks straightforward.

/lgtm

PR #75, opened 2021-06-30
Server unit tests

JoshVanL:
Adds unit tests to Server.

Branched from #74  

Adds an interface to certmanager for faking.

/assign @irbekrm 
/hold

### DIFF
--- a/pkg/certmanager/certmanager.go
+++ b/pkg/certmanager/certmanager.go
@@ -50,7 +50,7 @@ type Options struct {
 	IssuerRef cmmeta.ObjectReference
 }
 
-type Interface interface {
+type Signer interface {
 	// Sign will create a CertificateRequest based on the provided inputs. It will
 	// wait for it to reach a terminal state, before optionally deleting it if
 	// preserving CertificateRequests if turned off. Will return the certificate
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -66,7 +66,7 @@ type Server struct {
 
 	auther security.Authenticator
 
-	cm  certmanager.Interface
+	cm  certmanager.Signer
 	tls *tls.Provider
 
 	ready bool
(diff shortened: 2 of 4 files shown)


pkg/certmanager/certmanager.go
```diff
@@ -50,10 +50,18 @@ type Options struct {
 	IssuerRef cmmeta.ObjectReference
 }
 
-// Manager is used for signing CSRs via cert-manager. Manager will create
+type Interface interface {
```
irbekrm, line 53: Suggestion: call this `Signer` rather than `Interface`

irbekrm (changes requested):
The tests look great to me!
I've added an optional suggestion, otherwise happy to lgtm once #74 gets merged and this is rebased.

pkg/certmanager/certmanager.go
JoshVanL: Thanks @irbekrm. 

This is actually an intentional change. I've come around to the idea that `signer.Signer` is bad because it is stuttering, whereas `signer.Interface` makes sense and doesn't stutter.

irbekrm: I guess `certmanager.Signer` would not be a stutter and may be expressing the intent more clearly, but `Interface` is ok too imo

JoshVanL: This is true, I will change it to ‘Signer’.

JoshVanL:
Thanks @irbekrm, this is good for another review.

/hold cancel

pkg/certmanager/certmanager.go
JoshVanL: Done in https://github.com/cert-manager/istio-csr/pull/75/commits/8550a4c187ff733d02423553413fdbefba4cdce8

irbekrm:
@JoshVanL was this PR not supposed to have another commit with the server tests and the fake?

JoshVanL:
> @JoshVanL was this PR not supposed to have another commit with the server tests and the fake?

Not sure what's happened, they've managed to be apart of [this PR](https://github.com/cert-manager/istio-csr/pull/74) now. Strange.

irbekrm:
Ahh cool, so this is just the rename.
That looks good to me :) 

/lgtm

PR #79, opened 2021-07-01
Updates image tags to v0.2.0

JoshVanL:

/hold


irbekrm:
There must be some issue with the workers cluster, perhaps Calico or dns, we probably don't see this as much with `cert-manager` because of caching 

/test pull-istio-csr-k8s-v1-21-istio-v1-9

PR #82, opened 2021-07-27
Adds .app.tls.istiodCertificateDuration helm field as value to istiod's certificate.spec.,duration

### LINKED ISSUE #81: istiod certificate duration too short
Using AWS PCA the minimum cert duration is 24h and the istiod certificate is hard coded to 1h in the chart. Can this be made configurable in the values.yaml file?

https://github.com/cert-manager/istio-csr/blob/master/deploy/charts/istio-csr/templates/certificate.yaml#L16

JoshVanL:
This PR enables setting the istiod's `certificate.spec.duration` field using the helm field `app.tls.istiodCertificateDuration`.

The Helm version has been increased to prepare a new Helm Chart release. 

> Note that this is the first time the Helm Chart version will diversion from the `istio-csr` image version.

/assign @irbekrm 
fixes #81

### DIFF
--- a/deploy/charts/istio-csr/Chart.yaml
+++ b/deploy/charts/istio-csr/Chart.yaml
@@ -13,4 +13,4 @@ sources:
 - https://github.com/cert-manager/istio-csr
 
 appVersion: v0.2.0
-version: v0.2.0
+version: v0.2.1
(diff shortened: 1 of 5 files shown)


irbekrm:
I never looked at those DNS or Calico failures in workers' clusters that I said I'd look into :(

/retest

This looks good and non-controversial
/lgtm

irbekrm:
Actually I think the tests might be failing because we're picking up this 20s value from https://github.com/cert-manager/istio-csr/blob/master/hack/demo/istio-csr-values.yaml#L26 for the istiod cert.

JoshVanL:
Thanks for the catch @irbekrm, I've gone ahead and created a new helm field so that there is no conflict. I would like to keep that as 20s as a soft test in e2e.

JoshVanL:
@irbekrm Need another one :sweat_smile:

PR #88, opened 2021-08-23
Allows istio revisions to be configurable from helm chart

### LINKED ISSUE #86: Feature Request: Support Multiple Revisions
Due to the explicit certificate naming in the [helm chart](https://github.com/cert-manager/istio-csr/blob/master/deploy/charts/istio-csr/templates/certificate.yaml), istio-csr's helm chart doesn't support non-default revisions. 

Are there any plans to remediate this? One approach that comes to mind would be to accept an array of revisions in the helm chart (with the option to exclude the default revision) and loop over that value in the `certificate.yaml` template.

JoshVanL:

This PR expose the `app.istio.revisions` helm variable which modifies the DNS values which are present on the istiod Certificate. This allows for istio meshes which are running multiple revisions to still serve istiod correctly.

A new helm chart should be released once this is merged.

/assign @irbekrm 
/cc @keithmattix 
fixes #86

### DIFF
--- a/deploy/charts/istio-csr/templates/certificate.yaml
+++ b/deploy/charts/istio-csr/templates/certificate.yaml
@@ -5,9 +5,18 @@ metadata:
   namespace: istio-system
 spec:
   dnsNames:
+  {{- if .Values.app.istio.revisions }}
+  {{- $revisions := dict -}}
+  {{- range $revision := .Values.app.istio.revisions -}}
+  {{- if eq $revision "default" }}{{ $revision = "" }}{{ end }}
+  {{- if ne $revision "" }}{{ $revision = print "-" $revision }}{{ end }}
+  - istiod{{$revision}}.istio-system.svc
+  {{- end }}
+  {{- else }}
   - istiod.istio-system.svc
+  {{- end }}
   uris:
-    - spiffe://cluster.local/ns/istio-system/sa/istiod-service-account
+  - spiffe://{{.Values.app.tls.trustDomain}}/ns/istio-system/sa/istiod-service-account
   secretName: istiod-tls
   # Here we use a duration of 1 hour by default based on NIST 800-204A
   # recommendations (SM-DR13).
--- a/deploy/charts/istio-csr/values.yaml
+++ b/deploy/charts/istio-csr/values.yaml
@@ -99,6 +99,10 @@ app:
       # -- Container port to serve istio-csr gRPC service.
       port: 6443
 
+  istio:
+    # -- The istio revisions that are currently installed in the cluster.
+    revisions: []
+
   controller:
     leaderElectionNamespace: istio-system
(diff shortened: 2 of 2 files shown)


deploy/charts/istio-csr/templates/certificate.yaml
```diff
@@ -5,9 +5,18 @@ metadata:
   namespace: istio-system
 spec:
   dnsNames:
+  {{- if .Values.app.istio.revisions }}
+  {{- $revisions := dict -}}
```
keithmattix, line 9: Is there any need for this dict here?

keithmattix (approved):
Looks great! Thanks for the quick turnaround

deploy/charts/istio-csr/templates/certificate.yaml
JoshVanL: Thanks @keithmattix, looks like I ended up having some lines go missing. Should be better now with proper duplicate checking.

keithmattix: Excellent - thanks for adding that fix. LGTM!

keithmattix (approved):

deploy/charts/istio-csr/values.yaml
```diff
@@ -99,6 +99,10 @@ app:
       # -- Container port to serve istio-csr gRPC service.
       port: 6443
 
+  istio:
+    # -- The istio revisions that are currently installed in the cluster.
+    revisions: []
```
irbekrm, line 104: Suggestion: I guess typically deploying a new revision would be a day 2 operation and the users would likely first deploy Istio with a single control plane and at some point later decide to deploy a new revision.
So, it's likely that `istio-csr` would first be installed without `revisions` explicitly set. Then if they wanted to add a new revision 'canary', they would actually need to set this value to `["canary","default"] else the cert would only be created for `istiod-canary.istio-system.svc` DNS name.

Would it make sense to explicitly default `revisions` to ["default"]?

irbekrm (approved):
Looks good to me, I didn't know about the Istio revisions so was interesting to learn.

I've added one optional suggestion to default the revisions, so adding a hold for if you think that'd make sense, otherwise feel free to unhold.

/hold
/lgtm

deploy/charts/istio-csr/values.yaml
keithmattix: +1 Great point. I think that's a good default

JoshVanL: Yes, I thought about this however "default" is actually a [special revision case](https://istio.io/latest/docs/setup/upgrade/canary/#default-tag) which is different from not setting a revision.

It is my understanding that one need to actually setup their revision from `nothing` -> `default` -> `default + foo`. That assumption might be wrong though.

Issue #89, opened 2021-08-25
Support ARM64 images


xUnholy:
Is there any scope for adding ARM64 compatibly? People using Graviton 2 instances, or running on other ARM SBC devices will run into this architecture issue.  You could use docker buildx to build multi-architecture images.

```
standard_init_linux.go:228: exec user process caused: exec format error
```

xUnholy:
@JoshVanL did you need any help? I have some arm64 nodes I can test out any images you're building out at present.

JoshVanL:
Hi @xUnholy, apologies for the delay. I have opened a PR and will push the image shortly after it merges.

JoshVanL:
@xUnholy A multi-arch image and new helm chart version has been pushed. Let me know if things are working for you.

xUnholy:
Testing it out now. Here was the below findings:

```
  Warning  Unhealthy  2m39s (x6499 over 11h)  kubelet  Readiness probe failed: HTTP probe failed with statuscode: 500
```

```
panic: interface conversion: runtime.Object is nil, not *v1.CertificateRequest

goroutine 280 [running]:
github.com/cert-manager/istio-csr/pkg/certmanager.(*manager).waitForCertificateRequest(0x40002c8230, {0x26757d0, 0x400101ad40}, {0x26988f8, 0x4000cbc1e0}, 0x400037bb00)
	/workspace/pkg/certmanager/certmanager.go:188 +0x5b8
github.com/cert-manager/istio-csr/pkg/certmanager.(*manager).Sign(0x40002c8230, {0x26757d0, 0x400101ad40}, {0x22ff29c, 0x11}, {0x4001294000, 0x3c8, 0x799}, 0x34630b8a000, {0x4001048080, ...})
	/workspace/pkg/certmanager/certmanager.go:138 +0x550
github.com/cert-manager/istio-csr/pkg/tls.(*Provider).fetchCertificate(0x4000fcdea0, {0x26757d0, 0x400101ad40})
	/workspace/pkg/tls/tls.go:266 +0x284
github.com/cert-manager/istio-csr/pkg/tls.(*Provider).Start(0x4000fcdea0, {0x26757d0, 0x400101ad40})
	/workspace/pkg/tls/tls.go:119 +0x4c
sigs.k8s.io/controller-runtime/pkg/manager.(*controllerManager).startRunnable.func1(0x4001012000, {0x25f74a0, 0x4000fcdea0})
[...]

PR #95, opened 2021-09-15
Ensure that the watcher returns a non-nil object before decoding

JoshVanL:
In some strange circumstances the watcher may return a nil object, as should [here](https://github.com/cert-manager/istio-csr/issues/89#issuecomment-913235885).

This PR fixes that by ensuring the watcher object is not nil before decoding.

/assign @irbekrm 
/release-note-none
/kind bug

### DIFF
--- a/pkg/certmanager/certmanager.go
+++ b/pkg/certmanager/certmanager.go
@@ -184,10 +184,19 @@ func (m *manager) waitForCertificateRequest(ctx context.Context, log logr.Logger
 
 		log.V(3).Info("waiting for CertificateRequest to become ready")
 
-		w := <-watcher.ResultChan()
-		cr = w.Object.(*cmapi.CertificateRequest)
-		if w.Type == watch.Deleted {
-			return cr, errors.New("created CertificateRequest has been unexpectedly deleted")
+		for {
+			w := <-watcher.ResultChan()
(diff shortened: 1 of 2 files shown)


irbekrm:
assuming the usual

/retest

pkg/certmanager/certmanager.go
```diff
@@ -184,7 +184,11 @@ func (m *manager) waitForCertificateRequest(ctx context.Context, log logr.Logger
 
 		log.V(3).Info("waiting for CertificateRequest to become ready")
 
-		w := <-watcher.ResultChan()
+		var w watch.Event
+		for w.Object == nil {
```
irbekrm, line 188: Suggestion: instead of doing this couldn't you just do
```
cr, ok  := w.Object.(*cmapi.CertificateRequest)
if !ok {
  continue
}
```

That would I guess catch the nil edge case as well as any other where the object wouldn't be a CR

JoshVanL: Thanks, this is a much better way of doing it :+1:

irbekrm (approved):
Looks good to me 👍🏼 

/lgtm

irbekrm:
/retest

Or maybe needs to be rebased off the PR that creates docker network

PR #96, opened 2021-09-16
Certificate chain parsing and Root CA concatanation

JoshVanL:
This PR changes the tls provider so it will now expose a *x509.CertPool to consumers which contains the CertPool implementation of the current root CAs.

The main change of this PR is to update the server to now 1. ensure the certificate chain returned from the issuer has been issued by a trust root, and 2. append all current roots to the certificate response chain to the istio workload.

/assign @irbekrm 

```release-note
NONE
```


irbekrm (commented):
I've read through the code and this seems like a good and non-controversial change 👍🏼 

/lgtm
/approve

PR #98, opened 2021-09-17
Change context root to use controller-runtime signal handler

JoshVanL:
Replace our signal handler with the fuller featured controller-runtime one.

/kind cleanup
/assign @irbekrm 

```release-note
NONE
```


irbekrm (approved):
This is nice 👍🏼 Something to adapt in other projects too

PR #99, opened 2021-09-17
Test CA Rotation

JoshVanL:
Branched from #96 which should be merged first.

This PR adds a test for verifying CA rotation works with istio-csr.

The strategy is to install istio-csr with 2 root CAs. Initially the mesh and workloads are brought up signed via the first CA. The issuer is then swapped to the second, and we test that 1 workload can still talk to the other workload on another CA as they both trust both.

This need this<TODO> testing PR before merging.

/hold
/assign @irbekrm 

```release-note
NONE
```


Makefile
```diff
@@ -65,6 +65,9 @@ e2e: demo ## build demo cluster and runs end to end tests
 	./hack/run-e2e.sh
 	./hack/demo/destroy-demo.sh
 
+carotation: depend ## run ca rotation test
```
irbekrm, line 68: Nit: add to the list of phony targets

test/carotation/run.sh
```diff
@@ -0,0 +1,37 @@
...
+
+$TEST_DIR/setup-2.sh
+
+$TEST_DIR/setup-3.sh
+
+$TEST_DIR/test-1.sh
+
+$TEST_DIR/test-2.sh
+
+$TEST_DIR/cleanup-1.sh
```
irbekrm, line 37: Suggestion: [trap](https://man7.org/linux/man-pages/man1/trap.1p.html) this and move it to the top so that it runs also when any of the previous step fail

irbekrm (changes requested):
Looks great to me, I've added two very minor nits, happy to approve otherwise

irbekrm:
This is a genuine error, we don't have `jq` in the test image https://github.com/cert-manager/istio-csr/pull/99/commits/629d3852363639619f15cdc412ce1844930cbc17#diff-1fb42fb08fdc1a77ebb8c3f09fbb8f62777241ece55fa0191150965b08fd61dfR39

(I wonder if it wouldn't make sense to just build a test image with all these tools built-in)

PR #100, opened 2021-09-17
Root CA file watcher

JoshVanL:
Branched from #99  which should be merged first.

This PR introduces the `rootca` tls subpackage. If a root CAs file is defined, this package will watch for changes to this file and broadcast when it has changed. This is then used by the server and controller to distribute the changed CA roots without needed a  full restart.

/assign @irbekrm 

```release-note
NONE
```

### DIFF
--- a/pkg/controller/configmap.go
+++ b/pkg/controller/configmap.go
@@ -111,7 +111,7 @@ func AddConfigMapController(ctx context.Context, log logr.Logger, opts Options) 
 
 		// If the CA roots change then reconcile all ConfigMaps
 		Watches(&source.Channel{Source: c.tls.SubscribeRootCAsEvent()}, handler.EnqueueRequestsFromMapFunc(
-			func(obj client.Object) []reconcile.Request {
+			func(_ client.Object) []reconcile.Request {
 				var namespaceList corev1.NamespaceList
 				if err := c.lister.List(ctx, &namespaceList); err != nil {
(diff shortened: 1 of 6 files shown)


irbekrm:
Generally this looks good to me, but I am curious why the fsnotify did not work?

I've previously used [weaveworks/Watch](https://github.com/weaveworks/Watch) for watching changes to mounted configmaps etc like [they do in Prometheus](https://www.weave.works/blog/prometheus-configmaps-continuous-deployment/) (this may be out of date) and that uses fsnotify under the hood.

Apart from that, happy to lgtm 👍🏼

JoshVanL:
> Generally this looks good to me, but I am curious why the fsnotify did not work?
> 
> I've previously used [weaveworks/Watch](https://github.com/weaveworks/Watch) for watching changes to mounted configmaps etc like [they do in Prometheus](https://www.weave.works/blog/prometheus-configmaps-continuous-deployment/) (this may be out of date) and that uses fsnotify under the hood.
> 
> Apart from that, happy to lgtm 👍🏼

Thanks @irbekrm, I think the thing that tripped me up was that it takes a while for the file to actually change when the Secret is updated so my impatience testing fsnotify initially gave me the wrong impression.

I'm now using fsnotify which is more clean.

/assign @irbekrm

irbekrm (approved):
Looks good to me 👍🏼 

/lgtm

Issue #103, opened 2021-09-21
Not supplying static root CA "effectively making istio-csr TOFU"


joshmue:
First, thanks for this project! It glues together istio and cert-manager really well.

I am curious about the recent change in the README.md: It now highly recommends to explicitly supply a static root CA (PR #101):

> It is highly recommended that the root CA certificates are statically defined in istio-csr. If they are not, istio-csr will "discover" the root CA certificates when requesting its serving certificate, effectively making istio-csr TOFU.

I am not sure though, how the loading of the root CA cert from the serving certificate request's `CertificateRequest.status` would make istio-csr TOFU.

- When supplying it statically, istio-csr would trust K8s to supply the correct root CA file.  
- When not supplying it statically, it would rely on the configured K8s RBAC to prevent anything other than the cert-manager or other trusted components to make changes to `CertificateRequest` resources. At least in my understanding, this does not mean TOFU, as trust was already established via K8s RBAC mechanisms and service account authentication.

So, as far as I understand, both methods should be reasonably secure, granted RBAC is configured well.

Am I missing something here? Is the goal to protect against a compromised cert-manager?

JoshVanL:
Hi @joshmue, thanks for the kind words! :slightly_smiling_face: 

I think everything you said here is correct, and perhaps the wording ("effectively") could be changed here to be less alarming..

Indeed, the idea here is we are trying to protect against a malicious actor who, by some means, is able to highjack signing (update permissions on CertificateRequests) for the istio-csr signing service and workload's issuer. This could be a stolen token from an existing signer (cert-manager-controller is one), or by other means. With this ability, the actor could completely change or inject an additional CA which would be a vector for attack.

By statically defining the root CAs, this attack doesn't work. An attacker would only be able to stop workloads trusting istio-csr's service or prevent workloads from joining the mesh- a safe failure state. Of course one could argue that this is just moving the problem, however I think that there are some good volume options out there which have arguably better security profiles (the [secrets-store](https://github.com/kubernetes-sigs/secrets-store-csi-driver) csi driver backed by Vault or a cloud provider paired with workload identity comes to mind).

Whilst having the behavior of changing root CAs over time on the fly isn't necessarily _bad_, root CAs should be fairly static things and their delivery pipeline tightly controlled.

joshmue:
Hi @JoshVanL, thank you very much for the rapid and detailed response! This really clears this topic up for me.

[...]

PR #104, opened 2021-09-21
Update the wording around "TOFU" to not confuse users, and give a better explanation.

JoshVanL:
Related #103 

This PR removes the use of "TOFU" as the explanation as to why using a static root CA is preferred. It was brought up in the above issue that the use of Trust On First Use wasn't necessarily correct and it confuses the intent on why defining a static CA is better from a security standpoint.

I've also added a line giving inspiration of other volume sources that could be used.

/cc @joshmue
/assign @irbekrm 

```release-notes
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,8 +50,11 @@ ClusterIssuer installed earlier.
 
 > ⚠️ It is highly recommended that the root CA certificates are statically
 > defined in istio-csr. If they are not, istio-csr will "discover" the root CA
-> certificates when requesting its serving certificate, effectively making
-> istio-csr [TOFU](https://en.wikipedia.org/wiki/Trust_on_first_use).
+> certificates when requesting its serving certificate. Although discovering the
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -50,8 +50,11 @@ ClusterIssuer installed earlier.
 
 > ⚠️ It is highly recommended that the root CA certificates are statically
 > defined in istio-csr. If they are not, istio-csr will "discover" the root CA
-> certificates when requesting its serving certificate, effectively making
-> istio-csr [TOFU](https://en.wikipedia.org/wiki/Trust_on_first_use).
+> certificates when requesting its serving certificate. Although discovering the
+> root CAs reduces operational complexity, using CA pinning with a static bundle
+> is less venerable to
```
joshmue, line 55: Spotted a typo:

```suggestion
> is less vulnerable to
```

irbekrm (approved):
Looks good, nice to see users getting involved

/lgtm

PR #114, opened 2021-11-26
A service isn't a requirement for the servicemonitor

MattiasGees:
We shouldn't enforce the creation of the metrics service for the servicemonitor. The servicemonitor also picks up the default istio-csr service and when the metrics service is created, the metrics get picked up twice.

You can this in the following example on my cluster. I have 2 replicas and they get picked up twice
Screenshot from 2021-11-26 09-14-17

### DIFF
--- a/deploy/charts/istio-csr/templates/metrics-servicemonitor.yaml
+++ b/deploy/charts/istio-csr/templates/metrics-servicemonitor.yaml
@@ -1,4 +1,4 @@
-{{- if and .Values.app.metrics.service.enabled .Values.app.metrics.service.servicemonitor.enabled }}
+{{- if .Values.app.metrics.service.servicemonitor.enabled }}
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:


MattiasGees:
@JoshVanL additional, imo I think we should get rid of the service for metrics altogether and just use the default one. That will probably cause a breaking change in the values.yaml if we do this.

Currently
```
app:
  # -- Verbosity of istio-csr logging.
  logLevel: 1 # 1-5

  metrics:
    # -- Port for exposing Prometheus metrics on 0.0.0.0 on path '/metrics'.
    port: 9402
    # -- Service to expose metrics endpoint.
    service:
      # -- Create a Service resource to expose metrics endpoint.
      enabled: true
      # -- Service type to expose metrics.
      type: ClusterIP
      # -- ServiceMonitor resource for this Service.
      servicemonitor:
        enabled: false
        prometheusInstance: default
        interval: 10s
        scrapeTimeout: 5s
        labels: {}
```

What I think it will become if we remove the metrics service

```
app:
  # -- Verbosity of istio-csr logging.
  logLevel: 1 # 1-5

  metrics:
    # -- Port for exposing Prometheus metrics on 0.0.0.0 on path '/metrics'.
    port: 9402
    # -- ServiceMonitor resource for this Service.
    servicemonitor:
      enabled: false
      prometheusInstance: default
      interval: 10s
      scrapeTimeout: 5s
      labels: {}
```

JoshVanL:
Thanks, I agree we shouldn't _require_ that the metrics service is enabled for the service monitor. Though I don't think there is a need to remove the _option_ for a metrics service all together.

/lgtm
/approve
/ok-to-test

SgtCoDFish:
Haha, just realised this was LGTM'd exactly a year ago! Coincidence!

Thanks @MattiasGees :grin:

PR #116, opened 2021-12-10
Move "getting started" to entirely new document

SgtCoDFish:
The current instructions have several pain points which can lead to a failed
deployment which will be hard to debug.

The new guide should be simpler, easier to follow (and should actually work)

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,107 +17,84 @@
 
 istio-csr is an agent that allows for [Istio](https://istio.io) workload and
 control plane components to be secured using
-[cert-manager](https://cert-manager.io). Certificates facilitating mTLS, inter
-and intra cluster, will be signed, delivered and renewed using [cert-manager
+[cert-manager](https://cert-manager.io).
+
+Certificates facilitating mTLS &mdash; both inter
+and intra-cluster &mdash; will be signed, delivered and renewed using [cert-manager
 issuers](https://cert-manager.io/docs/concepts/issuer).
 
-⚠️ Currently supports Istio versions v1.7+
-
-⚠️ Currently supports cert-manager versions v1.3+
+istio-csr supports Istio v1.7+ and cert-manager v1.3+
 
 ---
 
-## Installation
+## Getting Started Guide For istio-csr
 
-### Installing cert-manager
+We have [a guide](./docs/getting_started.md) for setting up istio-csr in a fresh [kind](https://kind.sigs.k8s.io/docs/user/quick-start/#installation) cluster.
 
-Firstly, [cert-manager must be
-installed](https://cert-manager.io/docs/installation/) in your cluster.
+Following the guide is the best way to see istio-csr in action.
 
-### Issuer or ClusterIssuer
-An Issuer or ClusterIssuer must be configured which will be used to sign Istio
-certificates against.
+If you've already seen istio-csr in action or if you're experienced with running Istio and just want quick installation instructions, read on for more details.
 
-Here are examples of CA [Issuer](./docs/example-issuer.yaml) and
-[ClusterIssuer](./docs/example-cluster-issuer.yaml) configurations that are
-bootstrapped through self-signed issuers. It is advised to not use the CA Issuer
-type in production environments, and instead use an issuer who's CA private key
-material does not reside within the cluster.
+## Lower-Level Details (For Experienced Istio Users)
 
-> It is important to use an issuer type that is able to sign Istio mTLS workload
-> certificates (SPIFFE URI SANs) and
-> [istiod serving certificates](./deploy/charts/istio-csr/templates/certificate.yaml).
-> ACME issuers will not work.
+⚠️  The [getting started](./docs/getting_started.md) guide is a better place if you just want to try istio-csr out!
 
-If using an Issuer rather than the ClusterIssuer type, the Issuer must reside in
-the same namespace as that configured by `--certificate-namespace` on istio-csr,
-`istio-system` by default.
+Running istio-csr requires a few steps and preconditions in order:
 
-### Installing istio-csr
+1. A cluster _without_ Istio already installed
+2. cert-manager [installed](https://cert-manager.io/docs/installation/) in the cluster
+3. An `Issuer` or `ClusterIssuer` which will be used to issue Istio certificates
+4. istio-csr installed (likely via helm)
+5. Istio [installed](https://istio.io/latest/docs/setup/install/istioctl/) with some custom config required, e.g. using [the example config](./docs/istio-config-getting-started.yaml).
 
-Next, install istio-csr into the cluster, configured to use the Issuer or
-ClusterIssuer installed earlier.
+### Why Custom Istio Install Manifests?
 
-> ⚠️ It is highly recommended that the root CA certificates are statically
-> defined in istio-csr. If they are not, istio-csr will "discover" the root CA
-> certificates when requesting its serving certificate. Although discovering the
-> root CAs reduces operational complexity, using CA pinning with a static bundle
-> is less vulnerable to
-> [signer hijacking attacks](https://github.com/cert-manager/istio-csr/issues/103#issuecomment-923882792),
-> for example if a signer's token (such as cert-manager-controller) was stolen.
+If you take a look at the contents of [the example Istio install manifest](./docs/istio-config-getting-started.yaml) there are a few
+custom configuration options which are important.
 
-#### Discover root CAs installation
+Required changes include setting `ENABLE_CA_SERVER` to `false` and setting the `caAddress` from which Istio will
+request certificates; replacing the CA server is the whole point of istio-csr!
 
-```terminal
-$ helm repo add jetstack https://charts.jetstack.io
-$ helm repo update
-$ kubectl create namespace istio-system
-$ helm install -n cert-manager cert-manager-istio-csr jetstack/cert-manager-istio-csr
-```
+Mounting and statically specifying the root CA is also an important recommended step. Without a manually specified
+root CA istio-csr defaults to trying to discover root CAs automatically, which could theoretically lead to a
+[signer hijacking attack](https://github.com/cert-manager/istio-csr/issues/103#issuecomment-923882792) if for example
+a signer's token was stolen (such as the cert-manager controller's token).
 
-#### Load root CAs from file ca.pem (Preferred)
+### Issuer or ClusterIssuer?
 
-We use a Secret volume here to source the root CAs, however projects such as
-[secrets-store-csi-driver](https://github.com/kubernetes-sigs/secrets-store-csi-driver)
-are great options for improving security further.
+Unless you know you need a `ClusterIssuer` we'd recommend starting with an `Issuer`, since it should be easier to reason about
+the access controls for an Issuer; they're namespaced and so naturally a little more limited in scope.
 
-```terminal
-$ helm repo add jetstack https://charts.jetstack.io
-$ helm repo update
-$ kubectl create namespace istio-system
-$ kubectl create secret generic istio-root-ca --from-file=ca.pem=ca.pem -n cert-manager
-$ helm install -n cert-manager cert-manager-istio-csr jetstack/cert-manager-istio-csr \
-  --set "app.tls.rootCAFile=/var/run/secrets/istio-csr/ca.pem" \
-  --set "volumeMounts[0].name=root-ca" \
-  --set "volumeMounts[0].mountPath=/var/run/secrets/istio-csr" \
-  --set "volumes[0].name=root-ca" \
-  --set "volumes[0].secret.secretName=istio-root-ca"
-```
+That said, if you view your entire Kubernetes cluster as being a trust domain itself, then a ClusterIssuer is the more natural
+fit. The best choice will depend on your specific situation.
 
-All helm value options can be found [here](./deploy/charts/istio-csr/README.md).
+Our [getting started guide](./docs/getting_started.md) uses an `Issuer`.
 
-### Installing Istio
+### Which Issuer Type?
 
-If you are running Openshift, prepare the cluster for Istio.
-Follow instructions from Istio [platform setup
-guide](https://istio.io/latest/docs/setup/platform-setup/openshift/).
+Whether you choose to use an `Issuer` or a `ClusterIssuer`, you'll also need to choose the type of issuer you want such as:
 
-Finally, install Istio. Istio must be installed using the IstioOperator
-configuration changes within
-[`./hack/istio-config-x.yaml`](./hack/istio-config-1.11.4.yaml). Later versions
-of Istio share the same config.
+- [CA](https://cert-manager.io/docs/configuration/ca/)
+- [Vault](https://cert-manager.io/docs/configuration/vault/)
+- or an [external issuer](https://cert-manager.io/docs/configuration/external/)
 
-For OpenShift set the profile as `--set profile=openshift`.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -7,107 +7,20 @@
 
 istio-csr is an agent that allows for [Istio](https://istio.io) workload and
 control plane components to be secured using
-[cert-manager](https://cert-manager.io). Certificates facilitating mTLS, inter
-and intra cluster, will be signed, delivered and renewed using [cert-manager
+[cert-manager](https://cert-manager.io). Certificates facilitating mTLS - both inter
+and intra cluster - will be signed, delivered and renewed using [cert-manager
```
maelvls, line 11: I could not resist, sorry!! em-dashes 😅


as Michael Kashin would put it, "my life will never be the same."
<a href="https://twitter.com/networkop1/status/1465932140005269504"></a>

```suggestion
[cert-manager](https://cert-manager.io). Certificates facilitating mTLS — both inter
and intra-cluster — will be signed, delivered and renewed using [cert-manager
```

docs/getting_started.md
```diff
@@ -0,0 +1,114 @@
+# Getting Started With istio-csr
+
+This guide will run through installing and using istio-csr from scratch. We'll use [kind](https://kind.sigs.k8s.io/) to create a new cluster locally in Docker, but this guide should work on any cluster.
```
maelvls, line 3: There are extra steps that must be done so that the cluster can be ready before insalling Istio. The Istio docs call these setup tasks "Plaform Setup" (add policies on openshift, add firewall rules and such). Should we mention these "Platform Setup" steps here?

```suggestion
This guide will run through installing and using istio-csr from scratch. We'll use [kind](https://kind.sigs.k8s.io/) to create a new cluster locally in Docker, but this guide should work on any cluster as long as the [Platform Setup](https://istio.io/latest/docs/setup/platform-setup/) has been performed. Note that if you are following the Platform Setup for OpenShift, do not run the `istioctl install` command listed in the Istio documentation; we will be performing later in this getting started guide.
```

docs/getting_started.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Installing Istio
+
+If you're not running on kind, you may need to do some additional [setup tasks](https://istio.io/latest/docs/setup/platform-setup/) before installing Istio.
+
+This istio installation configuration is required for:
+
+- The CA server to be disabled in istiod
+- Ensuring Istio workloads request certificates from istio-csr
+- Ensuring that the istiod certificates and keys are mounted from the Certificate created when installing istio-csr
```
maelvls, line 107: I was a bit lost while reading here: what is the "installation configuration"?


```suggestion
We will be using the `istioctl` CLI to install Istio. In order for istio-csr to work, we will pass a custom IstioOperator manifest to `istioctl`. The custom manifest does the following:

- It disables the CA server in istiod,
- It makes sure that Istio workloads request certificates from istio-csr,
- It ensures that the istiod certificates and keys are mounted from the Certificate created when installing istio-csr.

Let us first download the custom IstioOperator manifest, and then use istioctl to install Istio:
```

maelvls (commented):
Great work!! Thank you! The guide is simpler to go through as someone new to Istio.

I successfully tested your new "getting started" tutorial on Kind and OpenShift. The OpenShift flow of instructions is a bit different than for other Kubernetes distributions, I suggested a few improvements to make the tutorial work with OpenShift.

docs/getting_started.md
```diff
@@ -0,0 +1,114 @@
...
+
+```console
+curl -sSL https://raw.githubusercontent.com/cert-manager/istio-csr/main/hack/istio-config-1.10.0.yaml > istio-install-config.yaml
+
+# This takes a little time to complete
+istioctl install -f istio-install-config.yaml
+```
```
maelvls, line 114: With OpenShift, the command varies slightly:

```sh
istioctl install --set profile=openshift -f <(curl -sSL https://raw.githubusercontent.com/SgtCoDFish/istio-csr/readmerewrite/hack/istio-config-1.10.0.yaml)
```

Could we somehow incorporate `--set profile=openshift` (for example, as a comment below the code block) for OpenShift users?

README.md
```diff
@@ -7,107 +7,20 @@
...
 issuers](https://cert-manager.io/docs/concepts/issuer).
 
-⚠️ Currently supports Istio versions v1.7+
-
-⚠️ Currently supports cert-manager versions v1.3+
+istio-csr supports Istio versions v1.7+ and cert-manager v1.3+
 
 ---
```
SgtCoDFish, line 17: self review after internal weekly meeting: README needs some high-level information for power uers

Also for most it'll likely be really hard in practice to install istio-csr _after_ istio, so we can also mention that in the README

docs/getting_started.md
SgtCoDFish: Good suggestion, thanks :+1:

docs/getting_started.md
SgtCoDFish: Much better wording yeah, thanks :+1:

docs/getting_started.md
SgtCoDFish: done, thanks :+1:

README.md
```diff
@@ -1,113 +1,79 @@
...
-$ kubectl create namespace istio-system
-$ kubectl create secret generic istio-root-ca --from-file=ca.pem=ca.pem -n cert-manager
-$ helm install -n cert-manager cert-manager-istio-csr jetstack/cert-manager-istio-csr \
-  --set "app.tls.rootCAFile=/var/run/secrets/istio-csr/ca.pem" \
-  --set "volumeMounts[0].name=root-ca" \
-  --set "volumeMounts[0].mountPath=/var/run/secrets/istio-csr" \
-  --set "volumes[0].name=root-ca" \
-  --set "volumes[0].secret.secretName=istio-root-ca"
-```
+Requried changes include setting `ENABLE_CA_SERVER` to `false` and setting the `caAddress` from which Istio will
```
maelvls, line 55:
```suggestion
Required changes include setting `ENABLE_CA_SERVER` to `false` and setting the `caAddress` from which Istio will
```

README.md
```diff
@@ -1,113 +1,79 @@
...
-Follow instructions from Istio [platform setup
-guide](https://istio.io/latest/docs/setup/platform-setup/openshift/).
+We'd recommend an `Issuer` unless you've got some reason to use a `ClusterIssuer`.
 
-Finally, install Istio. Istio must be installed using the IstioOperator
-configuration changes within
-[`./hack/istio-config-x.yaml`](./hack/istio-config-1.10.0.yaml). Later versions
-of Istio share the same config.
+It's much easier to reason about the access controls for an `Issuer`. In addition, `Issuer`s will ultimately require less work
+to configure, since our examples in the [getting started](./docs/getting_started.md) guide focus on `Issuer`s.
```
maelvls, line 67: Maybe we can add back the warning about the fact that the ACME issuer with Let's Encrypt is not supported in istio-csr due to the fact that the SPIFFE ID cannot be added to the X.509 subjectAltName field.

maelvls (dismissed):
Nice additions to the README. I like the fact that you made it clear why we have a "hack" manifest.

Except for the missing bit about the ACME issuer, I think we are good to go.

README.md
SgtCoDFish: Whoops, thanks :+1:

README.md
SgtCoDFish: Done, good idea again! :+1:

README.md
```diff
@@ -1,113 +1,96 @@
...
 
-### Installing Istio
+The key requirement is that arbitrary values can be placed into the `subjectAltName` (SAN) X.509 extension, since
+Istio places SPIFFE IDs there.
 
-If you are running Openshift, prepare the cluster for Istio.
-Follow instructions from Istio [platform setup
-guide](https://istio.io/latest/docs/setup/platform-setup/openshift/).
+That means that the ACME issuer **will not work** - publicly trusted certificates such as those issued by Let's Encrypt
+don't allow arbitrary entries in the SAN, for very good reasons.
```
maelvls, line 81:
```suggestion
That means that the ACME issuer **will not work** — publicly trusted certificates such as those issued by Let's Encrypt
don't allow arbitrary entries in the SAN, for very good reasons.
```

maelvls (approved):
Thanks for the PR!!

README.md
SgtCoDFish: Haha, fixed just because it's you! :wink:

docs/example-cluster-issuer.yaml
```diff
@@ -1,20 +1,32 @@
...
+# SelfSigned issuers are useful for creating root certificates
 apiVersion: cert-manager.io/v1
 kind: ClusterIssuer
 metadata:
   name: selfsigned
 spec:
   selfSigned: {}
 ---
+# Request a self-signed certificate from our ClusterIssuer; this will function as our
+# issuing root certifcate when we pass it into a CA ClusterIssuer.
```
jahrlin, line 13:
```suggestion
# issuing root certificate when we pass it into a CA ClusterIssuer.
```

docs/getting_started.md
```diff
@@ -0,0 +1,131 @@
...
+- Ensures that the istiod certificates and keys are mounted from the Certificate created when installing istio-csr.
+
+First we download our demo manifest and then we apply it.
+
+```console
+curl -sSL https://raw.githubusercontent.com/cert-manager/istio-csr/main/hack/istio-config-1.10.0.yaml > istio-install-config.yaml
+
+```
+
+You may wish to inspect and tweak `istio-isntall-config.yaml` if you know what you're doing,
```
jahrlin, line 119:
```suggestion
You may wish to inspect and tweak `istio-install-config.yaml` if you know what you're doing,
```

jahrlin (approved):
This is a great improvement, nice!

/lgtm

README.md
```diff
@@ -1,113 +1,96 @@
...
+If you've already seen istio-csr in action or if you're experienced with running Istio and just want quick installation instructions, read on for more details.
 
-Firstly, [cert-manager must be
-installed](https://cert-manager.io/docs/installation/) in your cluster.
+## Lower-Level Details (For Experienced Istio Users)
 
-### Issuer or ClusterIssuer
-An Issuer or ClusterIssuer must be configured which will be used to sign Istio
-certificates against.
+⚠️  The [getting started](./docs/getting_started.md) guide is a better place if you just want to try istio-csr out!
```
JoshVanL, line 40: Looks like this is repeated from 8 lines up.

README.md
```diff
@@ -1,113 +1,96 @@
...
-> ⚠️ It is highly recommended that the root CA certificates are statically
-> defined in istio-csr. If they are not, istio-csr will "discover" the root CA
-> certificates when requesting its serving certificate. Although discovering the
-> root CAs reduces operational complexity, using CA pinning with a static bundle
-> is less vulnerable to
-> [signer hijacking attacks](https://github.com/cert-manager/istio-csr/issues/103#issuecomment-923882792),
-> for example if a signer's token (such as cert-manager-controller) was stolen.
+Mounting and statically specifying the root CA is also an important recommended step. Without a manually specified
+root CA istio-csr defaults to trying to discover root CAs automatically, which could theoretically lead to a
+[signer hijacking attack](https://github.com/cert-manager/istio-csr/issues/103#issuecomment-923882792).
```
JoshVanL, line 60: This line is missing `for example if a signer's token (such as cert-manager-controller) was stolen.`

README.md
```diff
@@ -1,113 +1,96 @@
...
-#### Discover root CAs installation
+### Issuer or ClusterIssuer?
 
-```terminal
-$ helm repo add jetstack https://charts.jetstack.io
-$ helm repo update
-$ kubectl create namespace istio-system
-$ helm install -n cert-manager cert-manager-istio-csr jetstack/cert-manager-istio-csr
-```
+We'd recommend an `Issuer` unless you've got some reason to use a `ClusterIssuer`.
```
JoshVanL, line 64: Not sure this is necessarily correct, especially in the context of trust and the ideas for spiffe more generally (any workload should be able to request their identity from a trust domain).

hack/istio-config-1.10.0.yaml
```diff
@@ -7,7 +7,7 @@ spec:
   hub: gcr.io/istio-release
   meshConfig:
     # Change the following line to configure the trust domain of the Istio cluster.
-    trustDomain: foo.bar
+    trustDomain: cluster.local
```
JoshVanL, line 10: It's not clear to me why this is changing. We should either change back, or change every istio configuration file to this value.

JoshVanL (changes requested):
Couple comments from me :slightly_smiling_face: 

I wonder if we want to consider moving the getting started guide over to the cert-manager.io website for better visibility?

hack/istio-config-1.10.0.yaml
SgtCoDFish: > It's not clear to me why this is changing. 

My bad, probably should've commented on this one.

The reason for this is that the current guide in the README (i.e. the one on `main`) uses `hack/istio-config-1.10.0.yaml` verbatim for its example, but also suggests installing the helm chart whose default `app.tls.trustDomain` is `cluster.local`, which is one of the foot guns that everyone hit while doing the customer empathy session; i.e. this is one of the things that means if you follow the instructions you'll end up with a cluster that doesn't work.

> We should either change back, or change every istio configuration file to this value.

Yeah, fair point! I think probably either:

1. Change this back and set `app.tls.trustDomain` to `foo.bar` in the getting started guide
2. Provide `docs/istio-config-getting-started.yaml` for use in the getting started guide.

Given that the `hack/istio-config-x.yaml` files are used in the [e2e tests / demo](https://github.com/cert-manager/istio-csr/blob/ed48ecf4d305f0f970685bde1d23481ab0dc6940/hack/demo/deploy-demo.sh#L54), I think probably we want option 2, since we don't want tweaks to the e2e tests to affect our docs or vice versa. I'll make that change.

README.md
SgtCoDFish: This was intentional, I wanted it to be _super_ clear that the README isn't the place for newbies

README.md
SgtCoDFish: Yeah good addition, I'll do that :+1:

README.md
JoshVanL: (More of a general comment btw, don't want to block the on this point :slightly_smiling_face: )

SgtCoDFish: How about if it's a little more general:

> Unless you know you need a `ClusterIssuer` we'd recommend starting with an `Issuer`, since it should be easier to reason about the access controls for an `Issuer`; they're namespaced and so naturally a little more limited in scope.

> That said, if you view your entire Kubernetes cluster as being a trust domain itself, then a `ClusterIssuer` is a more natural fit. The best choice will depend on your specific situation.

Or if you disagree, any suggested language would be appreciated here!

JoshVanL: Yeah sounds great to me :+1:

JoshVanL:
@SgtCoDFish Once the config change has been reverted the tests will start passing. Guess there is now a conflict in the readme because of the 10 -> 11 change.

SgtCoDFish:
I've made the changes we discussed @JoshVanL and I've also updated the maintainer + description in the helm chart :+1:

SgtCoDFish:
Cheers all!

I realised I never addressed this:

> I wonder if we want to consider moving the getting started guide over to the cert-manager.io website for better visibility?

I think having a "subprojects" section on the site with some descriptions of various projects would be great in future, and this guide would be a good fit there too, but it's also IMO worth having the guide here in the README (and having this guide be the source of truth)!

PR #124, opened 2022-01-24
docs: Add validation steps to getting started guide

hawksight:
I found it difficult to determine if my setup was correct. After some learning I decided to add validation steps where I could see istio-csr in use. Adding the steps here as generically as possible with istio's example applications.

Signed-off-by: Peter Fiddes <peter.fiddes@gmail.com>


JoshVanL:
Thanks for the addition @hawksight 

/lgtm
/approve
/ok-to-test

Issue #125, opened 2022-01-24
Allow configuring filter for namespaces to create the RootCA ConfigMap


alandiegosantos:
istio-csr starts a control-loop that creates a ConfigMap _istio-ca-root-cert_ containing the Root Certificate used by istio-proxies. 
Not all namespaces requires this ConfigMap since no istio-proxy is deployed there or the Root CA bundle is distributed in a different manner.

Given this context, a parameter to allow filtering the namespaces that receives the configmap, similar to _istio-injection=enabled_ used by Istio to trigger the MutationWebHook.

JoshVanL:
Fixed in #126 
/close

PR #126, opened 2022-01-24
Allow filtering namespaces to create the ConfigMap istio-ca-root-cert

alandiegosantos:
This MR adds a parameter to filter namespaces in which the ConfigMap istio-ca-root-cert is created.
This parameter is validate on Complete() in [options.go](cmd/app/options/options.go) and propagated to
the [controller](pkg/controller/configmap.go).

So, the user can configure istio-csr to create the istio-root-ca-cert ConfigMap only on namespaces labelled _istio-csr-injection=enabled_ by adding the command-line parameter _--namespace-selector=istio-csr-injection=enabled_

### DIFF
--- a/cmd/app/options/options.go
+++ b/cmd/app/options/options.go
@@ -24,7 +24,6 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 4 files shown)


alandiegosantos:
Closes #125

cmd/app/options/options.go
```diff
@@ -221,4 +225,9 @@ func (o *Options) addControllerFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&o.Controller.LeaderElectionNamespace,
 		"leader-election-namespace", "istio-system",
 		"Namespace to use for controller leader election.")
+
+	fs.StringVar(&o.Controller.NamespaceSelector,
+		"namespace-selector", "",
```
JoshVanL, line 230: Small nit: can we change `namespace-selector` to `configmap-namespace-selector` to make it more explicit what we are doing with these namespaces at glance?

JoshVanL (commented):
Thanks @alandiegosantos!

All looks good to me, just a nit on the flag name.

/approve
/ok-to-test

cmd/app/options/options.go
alandiegosantos: @JoshVanL The parameter is renamed. Thanks for the suggestion.

JoshVanL:
Cheers!
/lgtm

PR #142, opened 2022-03-29
Adds warning message for `preserveCertificateRequests` and resource consumption

JoshVanL:
It is not currently clear to end users the pit falls of enabling `--preserve-certificate-requests`. Enabling this flag for a period of time on clusters with a non-trivial number of workloads leads to large amounts of resource consumption (since CertificateRequest resources are left around, effecting both literal memory usage, as well as size of list requests etc.). This resource consumption can eventually lead to ETCD, API server, and istio-csr errors and slow down. You really don't want to enable this flag in production.

I have updated the helm chart values comment, and CLI help output with a comment this effect. I have also gone ahead and added this warning comment to any values file where this option is present throughout the project to prevent anyone from missing it.

I have bumped the helm Chart version so we can publish a helm chart which includes this warning for end users.


anannaya (approved):

SgtCoDFish (approved):
/lgtm
/approve

Seems like a clear improvement, thanks :+1:

PR #150, opened 2022-05-03
Adds verify-helm-docs.sh and update-helm-docs.sh scripts. Verify helm-docs during lint

JoshVanL:
/assign @irbekrm 

This PR adds a scripts, run at make lint, which verifies that the helm chart README.md file is kept up to date.

An update script is also added which will update the helm chart README.md.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ test: lint ## test cert-manager-istio-csr
 	go test $$(go list ./pkg/... ./cmd/...)
 
 .PHONY: lint
-lint: boilerplate vet ## run code linting tests
+lint: boilerplate helm-docs vet ## run code linting tests
 
 .PHONY: vet
 vet:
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -112,3 +116,6 @@ $(BINDIR)/kubectl:
 
 $(BINDIR)/jq:
 	go build -o $(BINDIR)/jq github.com/itchyny/gojq/cmd/gojq
+
+$(BINDIR)/helm-docs:
+	go build -o $(BINDIR)/helm-docs github.com/norwoodj/helm-docs/cmd/helm-docs
```
irbekrm, line 121: Looks like a nice tool, I didn't know about this

irbekrm (approved):
Looks like a good addition, I didn't know about the docs tool!

/lgtm

PR #162, opened 2022-06-16
Adds hack config for istio v1.14.1 and changes default istio test version to v1.14.1

JoshVanL:
/assign @jahrlin


jakexks (approved):
/lgtm

PR #163, opened 2022-06-16
Create `go.mod` in `hack/tools`

JoshVanL:
Branched from #150 which should be merged first.

Creates a `go.mod` in `hack/tools` to track go modules there separately.

Also now build helm from source.

Also went ahead and changed `/bin/bash` to `/usr/bin/env bash` for hack scripts.

/assign @jahrlin


SgtCoDFish (approved):
/lgtm
/approve

Since this has already been lgtm'd I'll gladly approve it again to get things passing.

I think a more synchronized approach to this kind of thing across our projects would be great. It's a common pattern to install, say, helm, and we do it differently here than how we do it [in cert-manager](https://github.com/cert-manager/cert-manager/blob/a9c1e6aea73dea34d66280407649b61d58b79b3e/make/tools.mk#L156-L169)

Still, that yak doesn't need shaving now!

PR #165, opened 2022-07-11
Use a TokenRequest in token authentication tests

JoshVanL:
Now with Kubernetes v1.24, ServiceAccount tokens are now no longer stored in Kubernetes Secrets (whoop). As such, we can no longer rely on this method for generating tokens to test istio-csr's token authenticator. This PR now generates the token via the `TokenRequest` API.

This should get our e2e tests passing again.


irbekrm (approved):
I guess this change in 1.24 probably broke quite a lot of users (although it makes sense ofc).

/lgtm

PR #167, opened 2022-07-12
Increase bits to 4096 for more security

nitishkrishna:
Signed-off-by: Nitish Krishna Kaveri <nitish.krishna@getcruise.com>

### DIFF
--- a/cmd/app/options/options.go
+++ b/cmd/app/options/options.go
@@ -196,6 +196,10 @@ func (o *Options) addTLSFlags(fs *pflag.FlagSet) {
 		"A list of DNS names to request for the server's serving certificate which will be "+
 			"presented to istio-agents.")
 
+	fs.StringVar(&o.TLS.ServingCertificateKeySize,
+		"serving-certificate-key-size", "2048",
+		"Number of bits to use for the server's serving certificate (RSAKeySize).")
+
 }
 
 func (o *Options) addCertManagerFlags(fs *pflag.FlagSet) {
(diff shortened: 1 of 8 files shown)


JoshVanL (changes requested):
Thanks @nitishkrishna, couple comments from me!

nitishkrishna:
@JoshVanL Done! Thanks for the review!

pkg/server/internal/extensions/extensions_test.go
```diff
@@ -65,7 +65,7 @@ var (
 )
 
 func TestValidateCSRExtentions(t *testing.T) {
-	sk, err := pkiutil.GenerateRSAPrivateKey(2048)
+	sk, err := pkiutil.GenerateRSAPrivateKey(4096)
```
JoshVanL, line 68: It probably doesn't make sense for us to use `4096` key sizes in tests, since it is just going to slow things down for no real benefit.

cmd/app/options/options.go
```diff
@@ -196,6 +196,10 @@ func (o *Options) addTLSFlags(fs *pflag.FlagSet) {
 		"A list of DNS names to request for the server's serving certificate which will be "+
 			"presented to istio-agents.")
 
+	fs.StringVar(&o.TLS.ServingCertificateKeySize,
```
JoshVanL, line 199: Can we also expose this option in the helm chart? `app.server.serving.certificateKeySize` probably makes the most sense.

JoshVanL (changes requested):
Couple more things from me :slightly_smiling_face:

nitishkrishna:
@JoshVanL updated and added to chart README as well

JoshVanL:
Thanks @nitishkrishna! :slightly_smiling_face: 

/lgtm
/approve
/ok-to-test

JoshVanL:
@nitishkrishna looks like you need to run `./hack/update-helm-docs.sh` and commit the changes.

nitishkrishna:
@JoshVanL looks good now, sorry about that

PR #169, opened 2022-07-20
Remove istio config manifests in /hack for istio versions v1.7 to v1.9

JoshVanL:
The istio version v1.7 and v1.9 fail to install on Kubernetes v1.24 which is what we e2e test in.

See https://github.com/cert-manager/istio-csr/pull/163.

Rather than trying to play a dance with running different kube versions with different istio versions, let's just drop support for v1.9 and below istio-csr. v1.9 has been EOL since 2021-10 anyway.

/hold
Until we are ready to release so people who are using the current istio-csr version can still find the config.


SgtCoDFish:
From looking at [istio's supported releases](https://istio.io/latest/docs/releases/supported-releases/#support-status-of-istio-releases) the only version which actually explicitly supports + tests Kubernetes 1.24 is 1.14!

It looks like our test matrix is a little off if we're using 1.24 for everything?

I suppose that's a tangent for this PR though!

SgtCoDFish:
/override pull-cert-manager-istio-csr-k8s-v1-24-istio-v1-7 

(looks like this test was removed from [our tests](https://github.com/jetstack/testing/blob/master/config/jobs/cert-manager/istio-csr/cert-manager-istio-csr-presubmits.yaml) - maybe overriding will make it go away here)

SgtCoDFish:
/override pull-cert-manager-istio-csr-k8s-v1-24-istio-v1-8

(as above)

SgtCoDFish (approved):
/lgtm
/approve

Looks reasonable to do this; I see there's a hold which I'll leave in place :grin:

JoshVanL:
We're good to unhold this now :slightly_smiling_face: 

/unhold

PR #170, opened 2022-07-21
Use updated helm chart apiVersion for helm v3 support

nitishkrishna:
Signed-off-by: Nitish Krishna Kaveri <nitish.krishna@getcruise.com>

Trying to install istio-csr chart via [helmfile](https://github.com/helmfile/helmfile) results in the following linting error:
```
==> Linting /tmp/helmfile2642110333/cert-manager/cert-manager-istio-csr/cert-manager-istio-csr/cert-manager-istio-csr/v0.4.2/cert-manager-istio-csr
--
[ERROR] Chart.yaml: chart type is not valid in apiVersion 'v1'. It is valid in apiVersion 'v2'
```
From [docs](https://helm.sh/docs/topics/charts/#the-apiversion-field):
```
The apiVersion Field
The apiVersion field should be v2 for Helm charts that require at least Helm 3. Charts supporting previous Helm versions have an apiVersion set to v1 and are still installable by Helm 3.

Changes from v1 to v2:

A dependencies field defining chart dependencies, which were located in a separate requirements.yaml file for v1 charts (see Chart Dependencies).
The type field, discriminating application and library charts (see Chart Types).
```

### DIFF
--- a/deploy/charts/istio-csr/Chart.yaml
+++ b/deploy/charts/istio-csr/Chart.yaml
@@ -1,4 +1,4 @@
-apiVersion: v1
+apiVersion: v2
 
 name: cert-manager-istio-csr
 type: application


JoshVanL:
Thanks again @nitishkrishna! 

/ok-to-test
/lgtm
/approve

JoshVanL:
Once the current set of PRs have merged, I plan on making a new release so it is not worth increasing the version of the helm Chart IMO.

PR #171, opened 2022-07-22
Increase the kubectl wait timeout for the e2e test carotation setup scripts

JoshVanL:
It is often the case that the `carotation` e2e test scripts with fail will a "flake". This is due to resources not yet being in a Ready state in time when setting up the test. This happens particularly in CI, where there are larger resource constraints so everything running in kind is running a bit slower.

This PR increases the kubectl wait command timeouts from the default 30 seconds to 180 seconds. Whilst also being in line with other wait commands in the test suite, this change doesn't change the integrity of the test since these are just commands waiting for `cert-manager` (not `istio-csr`) to reconcile the test dependencies.


SgtCoDFish (approved):
/lgtm
/approve

Seems clear, thanks!

PR #179, opened 2022-09-02
Allow ECDSA serving certificate

Alkemic:
Why this change:
In our environment we are using ECDSA certificates only, unfortunately by design istio-csr uses only RSA. New flag `--serving-use-ec-algorithm` changes serving certificate to ECDSA, and using `app.server.serving.useECAlgo` in helm adds this flag to k8s deployment and changes private key for `istiod.istio-system.svc` certificate.


cmd/app/options/options.go
```diff
@@ -200,6 +200,8 @@ func (o *Options) addTLSFlags(fs *pflag.FlagSet) {
 		"serving-certificate-key-size", 2048,
 		"Number of bits to use for the server's serving certificate (RSAKeySize).")
 
+	fs.BoolVar(&o.TLS.ServingUseECAlgorithm, "serving-use-ec-algorithm", false,
+		"Use Elliptical Signature algorithm to use when generating private keys.")
```
JoshVanL, line 204: We should change this to be a string with accepts either `"RSA"` or `"ECDSA"` which defaults to `"RSA"`.

deploy/charts/istio-csr/templates/certificate.yaml
```diff
@@ -34,8 +34,13 @@ spec:
   renewBefore: {{ .Values.app.tls.istiodCertificateRenewBefore }}
   privateKey:
     rotationPolicy: Always
+    {{- if .Values.app.server.serving.useECAlgo }}
+    algorithm: ECDSA
+    size: 256
+    {{- else }}
     algorithm: RSA
     size: {{ .Values.app.tls.istiodPrivateKeySize }}
+    {{- end }}
```
JoshVanL, line 43:
```suggestion
    algorithm: {{ .Values.app.server.serving.algorithm }}
    size: {{ .Values.app.tls.istiodPrivateKeySize }}
```

JoshVanL (changes requested):
Thanks @Alkemic! Can we please make declaring the algorithm to use a string, and default everything to RSA (to keep the default the same behavior as today.

/assign @Alkemic

cmd/app/options/options.go
Alkemic: Done. Btw: do you think it's worth to add validation to one of two?

JoshVanL: Yes please!

SgtCoDFish:
I'd be willing to see this get over the line, but the test failures look legitimate!

inteon:
@SgtCoDFish This PR should be ready to be merged now.

SgtCoDFish (approved):
/lgtm
/approve

Thanks @Alkemic for raising this and @inteon for the final few tweaks!

I haven't tested this thoroughly, but the changes make sense to me in a vacuum so I'm happy to merge :+1:

Issue #180, opened 2022-10-03
Metrics scraped twice


konnoska:
Setting the following values:

```
app:
  metrics:
    service:
      enabled: true
      servicemonitor:
        enabled: true
```
Creates a servicemonitor that targets both `<istio-csr-servicename>-metrics` and `<istio-csr-servicename>` resulting in pod metrics being scraped twice.

PR #181, opened 2022-10-03
fix: metrics service labels

konnoska:
Servicemonitor should have a selector that only matches service-metrics and not the app-service, in order to avoind adding each prometheus target twice.

fixes: https://github.com/cert-manager/istio-csr/issues/180


JoshVanL:
Thanks!

/lgtm
/approve
/ok-to-test

PR #187, opened 2022-12-20
Add the ability to ignore cluster scoped resources. 

knutejohKLP:
These resources will generally be handled seperatly in a multitenant enviornment. This also gives the ability to control where the service account can create the configmaps, as clusterwide read/write on configmaps could be to broad access for security reasons.

Signed-off-by: Knut-Erik Johnsen <knut-erik.johnsen@klp.no>

### DIFF
--- a/deploy/charts/istio-csr/templates/clusterrole.yaml
+++ b/deploy/charts/istio-csr/templates/clusterrole.yaml
@@ -1,3 +1,4 @@
+{{- if not .Values.ignoreClusterResources }}
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
@@ -21,3 +22,4 @@ rules:
   - "tokenreviews"
   verbs:
   - "create"
+{{- end }}
--- a/deploy/charts/istio-csr/README.md
+++ b/deploy/charts/istio-csr/README.md
@@ -50,6 +50,7 @@ istio-csr enables the use of cert-manager for issuing certificates in Istio serv
 | app.tls.istiodPrivateKeySize | int | `2048` |  |
 | app.tls.rootCAFile | string | `nil` | An optional file location to a PEM encoded root CA that the root CA ConfigMap in all namespaces will be populated with. If empty, the CA returned from cert-manager for the serving certificate will be used. |
 | app.tls.trustDomain | string | `"cluster.local"` | The Istio cluster's trust domain. |
+| ignoreClusterResources | bool | `false` | Ignore creation of clusterscoped resources, ClusterRole and ClusterRolebinding. Useful in multitenant deployments. |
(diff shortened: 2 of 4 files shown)


knutejohKLP:
//cc @SgtCoDFish

SgtCoDFish:
Hey, I'm sorry but I don't have the bandwidth to review this right now. I'll raise it with the other maintainers and hopefully someone can take a look!

deploy/charts/istio-csr/templates/clusterrole.yaml
```diff
@@ -21,3 +22,4 @@ rules:
   - "tokenreviews"
```
munnerz, line 22: Does this actually work? istio-csr requires the ability to perform a TokenReview... so by not granting this permission, it surely won't be able to do _anything_?

knutejohKLP: The idea here is that the deployments on the cluster level, the role and the rolebinding, could be handled outside of this deployment. We are running multiple deployments of istio and istio-csr in our clusters, but these resources can only be set once, so we can't use the helmtemplates as they are right now. For a single deployment, this should always be created, which is why the default is false on the setting.

deploy/charts/istio-csr/templates/clusterrole.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if not .Values.ignoreClusterResources }}
 kind: ClusterRole
```
munnerz, line 2: For this namespaced approach, you are still going to need to grant these permissions on a per namespace basis right? So keeping this *ClusterRole* would allow users to create namespace-scoped RoleBindings in each namespace that they want enabling for istio-csr use.

Is there a reason to not deploy the ClusterRole? Otherwise this will lead to 1) duplicated resources (everyone needs to create a RoleBinding in every namespace) and 2) uncertainty, as users won't know what permissions to grant in those Roles.

knutejohKLP: If we leave the clusterrole, we can't use the helm charts for more than one deployment, as they will get the same name. The deployments will complain about resources that already exists. The idea is that anything on the clusterlevel is handled outside of this helm chart when running in a multitenant setup. and that the helm charts focus on a single namespace.

knutejohKLP:
Any more feedback on this issue?

SgtCoDFish:
Hey, this is obviously very old now and out of date. I'm sorry this got lost in the mix.

I'm going to close this since so much has changed. If you'd like to re-raise it, please feel free to and I hope we'll do better at reviewing it next time. We can be pinged in [kubernetes slack](https://cert-manager.io/docs/contributing/#slack) for reviews too, which is a great way to get maintainer eyeballs.

Sorry again!

PR #193, opened 2023-02-17
Update OWNERS file

inteon:
See https://www.kubernetes.dev/docs/guide/owners/ for more info.

Unfortunately, it is not possible to use a GH Teams in this OWNERS file:

> We use aliases for groups instead of GitHub Teams, because changes to GitHub Teams are not publicly auditable.
> - https://www.kubernetes.dev/docs/guide/owners/#owners_aliases

Instead, I created an OWNERS_ALIASES file that will have to be kept in-sync across all our repos.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,7 +1,4 @@
 approvers:
-- joshvanl
-- jakexks
-- SgtCoDFish
+- cm-maintainers
 reviewers:
-- joshvanl
-- SgtCoDFish
+- cm-maintainers
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -1,0 +1,11 @@
+aliases:
+  cm-maintainers:
+    - munnerz
+    - joshvanl
+    - meyskens
+    - wallrj
+    - jakexks
+    - maelvls
+    - irbekrm
+    - sgtcodfish
+    - inteon


SgtCoDFish:
> Instead, I created an OWNERS_ALIASES file that will have to be kept in-sync across all our repos.

It's a real shame that we can't use GH teams or other sources of teams.

I don't think having OWNERS_ALIASES is helpful to us here. We still have a file we need to sync across all repos when the maintainership group changes, just now it's called `OWNERS_ALIASES` rather than `OWNERS` and we have to check two files rather than one.

If we can't have some other external definition of who the core group of maintainers is, would it not be easier to just have OWNERS be used everywhere than to add the aliases?

inteon:
The advantage of OWNERS_ALIASES is that we have only one list `cm-maintainers` instead of `approvers` and `reviewers`. Also, it makes it easier to see who is added to the `approvers`/ `reviewers` on top of the list of  `cm-maintainers`.
eg.
```
approvers:
- cm-maintainers
reviewers:
- cm-maintainers
- user123
```
I do agree however that there are only minor advantages.

SgtCoDFish (approved):
/lgtm
/approve

I'm not convinced that having ALIASES is worth it from a maintenance standpoint just because the way we use these files in the cert-manager org is so radically different from the way it's used [in k/k](https://github.com/kubernetes/kubernetes/blob/c97512fe2a7ae4562a8a5d1daefac20d05a7e80d/OWNERS_ALIASES).

That said, I don't see a reason to block this because it'll work. Down the line we'll probably benefit from a unified approach, but for now let's at least get the list of maintainers for this repo updated!

PR #194, opened 2023-02-17
Add IstioOperator CR for v1.17.0

inteon:
With https://github.com/jetstack/testing/pull/886 merged, we now test istio 1.17.0!
This PR adds the configuration necessary for that change.


irbekrm:
@inteon actually I think they just had a mishap in their release process https://istio.slack.com/archives/C04FA3P3BU3/p1676480848476929

So maybe not worth changing the refs, perhaps we're faster pulling from gcr as we run tests in gcr?

SgtCoDFish (approved):
/lgtm
/approve

Thanks!

PR #195, opened 2023-02-19
Limit namespaces for configmaps, useful for multitenant environments

knutejohKLP:
When running in a multitenant environment, limit the namespaces the configmaps are written to be using the built in --configmap-namespace-selector option in istio-csr


SpectralHiss:
This seems to work fine, i just want to have another example in documentation for non-maistra use-case like perhaps `istio-injection=enabled` or `istio.io/rev=canary`

SpectralHiss:
I did notice that the reconciliation loop doesn't work well if the namespace was already created at the time of adding the namespace selector (an upgrade scenario), so i will double check that and open a separate issue if needed.

SpectralHiss:
@SgtCoDFish tested, waiting for your approval.

SgtCoDFish (approved):
/lgtm
/approve

Approving based on @SpectralHiss 's testing. Thanks very much for your contribution!

PR #196, opened 2023-02-21
Seperate Role permissions between leases for leader election and cert…

### LINKED ISSUE #191: istio-csr should seperate leases role permissions from cert-manager issuer namespace
The current behaviour creates the permissions for leases in the same namespace as the cert-manager issuer namespace.
This is an overlooked error when the leader election namespace != app.cert-manager namespace. 
likely to be encountered in multi-tenant environments and very confusing.

SpectralHiss:
Fixes #191 

* Fixes the configuration for the scenario where the leader election namespace is not the same as the cert-manager Issuer namespace.
* Most commonly arises when we are installing in a multi-tenant where Istio is not in istio-system namespace.
* For example, Istio can be installed with mesh1 as the default namespace, the issuer could still live in an istio-system or cert-manager namespace and the leader election namespace can be mesh1 or any other namespace if needed.

I've also checked and the events broadcaster thing needs to stay in the cert-manager namespace as it relates to certificate issuance/lifecycle events which are broadcast to that namespace by default.

### DIFF
--- a/deploy/charts/istio-csr/templates/role_leases.yaml
+++ b/deploy/charts/istio-csr/templates/role_leases.yaml
@@ -1,0 +1,19 @@
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  labels:
+{{ include "cert-manager-istio-csr.labels" . | indent 4 }}
+  name: {{ include "cert-manager-istio-csr.name" . }}-leases
+  namespace: {{ .Values.app.controller.leaderElectionNamespace }}
+rules:
+- apiGroups:
+  - "coordination.k8s.io"
+  resources:
+  - "leases"
+  verbs:
+  - "get"
+  - "create"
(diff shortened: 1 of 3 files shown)


deploy/charts/istio-csr/templates/role_leases.yaml
```diff
@@ -0,0 +1,19 @@
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  labels:
+{{ include "cert-manager-istio-csr.labels" . | indent 4 }}
+  name: {{ include "cert-manager-istio-csr.name" . }}-leases
```
inteon, line 6: We need an additional RoleBinding for this role

inteon (changes requested):

inteon (approved):
Thank you for fixing this bug!
/lgtm

PR #202, opened 2023-03-15
Support adding pod annotations

dblackdblack:
My logging provider requires that I annotate all pods with a routing key. This PR expands the helm chart to allow setting arbitrary annotations on the pods.


SgtCoDFish:
Hey, sorry this took a while to get eyes on. Closing in favour of #372

PR #210, opened 2023-05-31
Firefly policy annotation

SpectralHiss:
WIP PR

### DIFF
--- a/cmd/app/options/options.go
+++ b/cmd/app/options/options.go
@@ -136,6 +136,7 @@ func (o *Options) addFlags(cmd *cobra.Command) {
 	o.addTLSFlags(nfs.FlagSet("TLS"))
 	o.addServerFlags(nfs.FlagSet("Server"))
 	o.addControllerFlags(nfs.FlagSet("controller"))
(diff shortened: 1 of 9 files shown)


cmd/app/options/options.go
```diff
@@ -136,6 +136,7 @@ func (o *Options) addFlags(cmd *cobra.Command) {
 	o.addTLSFlags(nfs.FlagSet("TLS"))
 	o.addServerFlags(nfs.FlagSet("Server"))
 	o.addControllerFlags(nfs.FlagSet("controller"))
+	o.addFireflyFlags(nfs.FlagSet("firefly"))
```
SpectralHiss, line 139: we have to remove these right

JoshVanL: I don't think any vendor specific flags here are particularly appropriate. A '--certificate-request-annotations' map[string]string flag totally makes sense to me though. Vendors can then share their user friendly helm template options to configure for their software in their documentation. Even having a vendor example in the helm chart is probably fine.

Issue #208, opened 2023-05-11
Add ability to annotate certificate requests generateed by istio-csr


SpectralHiss:
The specific need is relating to integrating with a Venafi project "Firefly" which allows for rooting your mesh PKI in your enterprise root.
The ask could be useful more generally and definitely wouldn't tie istio-csr to Firefly.

Firefly has a comprehensive policy framework which allows you to apply policy for certificates issued by the mesh, however, certificaterequests need to be annotated with the particular policy that is enforced for the cluster identities for the right policy to take effect.

When istio-csr runs there are effectively two classes of certificates with different lifetime and purpose, there operational certificates such as the one for istiod server and istio-csr replicas as well as obviously all the mtls certificates and they all need to be signed by the same Issuer specified in Values file.

A single annotation that is included on all CertificateRequest specifying the policy enforced in the cluster would allow for Firefly to enact policy when issuing, but the policy needs to be broad enough to capture all classes of certificates.

It would be an even better refinement if istio-csr could have different annotations for these different "classes" of certificates
I'd like the ability to specify some annotations on the certificaterequests generated by istio-csr, this just needs to be a static one for now so setting a single one in the values file will do for a first iteration. unless there is a more elegant solution proposed by the cert-manager team.

SpectralHiss:
The above contribution by @aidy with input from me and @JoshVanL closes this issue, it is a first approximation where a single annotation is set to all workload certificates. Hopefully these annotations will be helpful to the wider community.

an example has been added to the repo https://github.com/aidy/istio-csr/blob/41d2249db6ea7f639fe17e0ddd2a4aa99d657b78/hack/demo/istio-csr-values.yaml#L24, a release will be cut soon.

PR #212, opened 2023-06-08
Add ability to specify extra annotations on CRs

aidy:
connects #208


internal/controller/feature/features.go
```diff
@@ -0,0 +1,43 @@
...
+
+	utilfeature "github.com/cert-manager/istio-csr/pkg/util/feature"
+)
+
+const (
+	// alpha: v0.7.2
+	//
+	// AdditionalAnnotations allows additional annotations to be specified on
+	// certificate requests
+	AdditionalAnnotations featuregate.Feature = "AdditionalAnnotations"
```
JoshVanL, line 31: Is it really necessary to make this feature gated? Seems like a no-brainer good feature to me. Are we concerned that the flag schema is going to change?

pkg/certmanager/certmanager.go
```diff
@@ -48,6 +51,8 @@ type Options struct {
 
 	// IssuerRef is used as the issuerRef on created CertificateRequests.
 	IssuerRef cmmeta.ObjectReference
+
+	AdditionalAnnotations map[string]string
```
JoshVanL, line 55: Add comment. Also add validation to ensure that all key values acceptable as annotations (https://github.com/kubernetes/apimachinery/blob/6b1428efc73348cc1c33935f3a39ab0f2f01d23d/pkg/api/validation/objectmeta.go#L44)

JoshVanL (changes requested):

internal/controller/feature/features.go
aidy: Yep - fair enough. We came at it initially from a different angle, but there's no real need for it.

pkg/certmanager/certmanager.go
JoshVanL: We just need to use that function linked, and return error and exit error the program if the given annotations aren't permissible.

aidy: Yes, I wasn't sure where to perform that exit, but I think I've found an appropriate place now.

cmd/app/app.go
```diff
@@ -52,6 +54,13 @@ func NewCommand(ctx context.Context) *cobra.Command {
 				return err
 			}
 
+			if errs := validation.ValidateAnnotations(opts.CertManager.AdditionalAnnotations, nil); len(errs) != 0 {
+				errStrings := []string{}
+				for _, err := range errs {
+					errStrings = append(errStrings, fmt.Errorf("Invalid annotation: %w", err).Error())
+				}
+				return fmt.Errorf(strings.Join(errStrings, "\n"))
+			}
```
JoshVanL, line 63:
```suggestion
			if errs := validation.ValidateAnnotations(opts.CertManager.AdditionalAnnotations, nil); len(errs) > 0 {
				return fmt.Errorf("invalid annotations: %w", errs.ToAggregate())
			}
```

cmd/app/options/options.go
```diff
@@ -228,6 +229,12 @@ func (o *Options) addCertManagerFlags(fs *pflag.FlagSet) {
 		"Group of the issuer to sign istio workload certificates.")
 }
 
+func (o *Options) addAdditionalAnnotationsFlags(fs *pflag.FlagSet) {
+	fs.StringToStringVar(&o.CertManager.AdditionalAnnotations,
+		"additional-annotations", map[string]string{},
+		"Any additional annotations to include on certificate requests")
```
JoshVanL, line 235:
```suggestion
		"Additional annotations to include on created CertificateRequests resources.")
```

JoshVanL (changes requested):

cmd/app/app.go
aidy: That's much neater - thanks!

cmd/app/app.go
```diff
@@ -52,6 +54,13 @@ func NewCommand(ctx context.Context) *cobra.Command {
 				return err
 			}
 
+			if errs := validation.ValidateAnnotations(opts.CertManager.AdditionalAnnotations, nil); len(errs) != 0 {
```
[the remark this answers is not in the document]
JoshVanL: If we do `field.NewPath("certificate-request-additional-annotations")` rather than nil, does it produce nicer output?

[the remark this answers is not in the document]
aidy:
```
Error: invalid annotations: certificate-request-additional-annotations: Invalid value: "_foo.baz": name part must consist of alphanumeric characters, '-'
, '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]')
```

I reckon we just return `errs.ToAggregate()` (no `fmt.Errorf`), so that we get:
```
Error: certificate-request-additional-annotations: Invalid value: "_foo.baz": name part must consist of alphanumeric characters, '-', '_' or '.', and mus
t start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]')
```

JoshVanL (changes requested):

pkg/certmanager/certmanager.go
```diff
@@ -108,6 +111,9 @@ func (m *manager) Sign(ctx context.Context, identities string, csrPEM []byte, du
 		},
 	}
 
+	for k, v := range m.opts.AdditionalAnnotations {
+		cr.ObjectMeta.Annotations[k] = v
+	}
```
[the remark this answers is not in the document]
aidy: Guaranteed not nil - https://github.com/cert-manager/istio-csr/pull/212/files#diff-a14578667edea6930bb709d1c7e49df6f71e2c2b07b37242751075d9322e10f3R100

deploy/charts/istio-csr/templates/deployment.yaml
```diff
@@ -58,6 +58,16 @@ spec:
...
           - "--preserve-certificate-requests={{.Values.app.certmanager.preserveCertificateRequests}}"
 
+            # AdditionalAnnotations
+          {{- if .Values.app.certmanager.additionalAnnotations }}
+            {{- $annotations := list }}
+            {{- range $annotation := .Values.app.certmanager.additionalAnnotations }}
+              {{- $x := list $annotation.name $annotation.value | join "=" }}
+              {{- $annotations = append $annotations $x }}
+            {{- end }}
+          - "--additional-annotations={{ join "," $annotations }}"
```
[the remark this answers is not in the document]
aidy: Good call - hadn't originally because it was featuregate'd. Done.

PR #220, opened 2023-11-05
Adds option in Makefile to check code coverage by unit tests

TrilokGeer:
A new option _make check-coverage_ is added to check code coverage by unit tests. It uses an optional _threshold_ parameter to validate based on required code coverage. The option helps in automation to mandate unit tests and required code coverage for maintaining code quality.


inteon:
@TrilokGeer thank you for your PR.
We added code coverage reporting as part of our refactor of the Makefiles: https://github.com/cert-manager/istio-csr/blob/main/make/test-unit.mk#L23.

PR #221, opened 2023-11-18
Migrate makefiles and CI/CD

inteon:
ref: https://github.com/cert-manager/makefile-modules/issues/3


make/_shared/cert-manager/01_mod.mk
```diff
@@ -0,0 +1,17 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+cert_manager_crds := $(bin_dir)/scratch/cert-manager-crds.yaml
+$(cert_manager_crds): | $(bin_dir)/scratch
+	curl -sSLo $@ https://github.com/cert-manager/cert-manager/releases/download/$(cert_manager_version)/cert-manager.crds.yaml
```
SgtCoDFish, line 17: suggestion (not a blocker for this PR but maybe something wider): If we don't include a version number in the file we create, it won't be recreated when we bump `cert_manager_version`.

Could we do something like this?

```suggestion
cert_manager_crds := $(bin_dir)/scratch/cert-manager-crds-$(cert_manager_version).yaml
$(cert_manager_crds): | $(bin_dir)/scratch
	curl -sSLo $@ https://github.com/cert-manager/cert-manager/releases/download/$(cert_manager_version)/cert-manager.crds.yaml
```

Of course this means that the location is no longer constant but we can always pass the value of `cert_manager_crds` as a variable, or create a new `$(bin_dir)/scratch/cert-manager-crds.yaml` target which depends on `$(bin_dir)/scratch/cert-manager-crds-$(cert_manager_version).yaml`

make/_shared/tools/00_mod.mk
```diff
@@ -0,0 +1,490 @@
...
+# https://github.com/helm/helm/releases
+TOOLS += helm=v3.13.3
+# https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl
+TOOLS += kubectl=v1.29.0
+# https://github.com/kubernetes-sigs/kind/releases
+TOOLS += kind=v0.20.0
+# https://www.vaultproject.io/downloads
+TOOLS += vault=1.15.4
+# https://github.com/Azure/azure-workload-identity/releases
+TOOLS += azwi=v1.2.0
```
SgtCoDFish, line 48: question: Where's azwi coming from? I can't see it as currently being a dependency of istio-csr? Same question for vault I think and maybe others.

deploy/charts/istio-csr/templates/deployment.yaml
```diff
@@ -45,7 +45,6 @@ spec:
             path: {{ .Values.app.readinessProbe.path }}
           initialDelaySeconds: 3
           periodSeconds: 7
-        command: ["cert-manager-istio-csr"]
```
SgtCoDFish, line 48: question: Why this change? Does the entrypoint change in the built container?

test/carotation/workloads/httpbin.yaml
```diff
@@ -63,7 +63,7 @@ spec:
     spec:
       serviceAccountName: httpbin
       containers:
-      - image: quay.io/joshvanl_jetstack/httpbin:latest
```
SgtCoDFish, line 66: comment: Yikes, good catch! I'd support moving this out of this PR and merging this immediately, but I'm also fine with including it here

SgtCoDFish (commented):
Got a few questions: nothing major, but I'd like to understand more! I don't see any real blockers though.

make/_shared/tools/00_mod.mk
inteon: This library is part of the make-file modules, so it shouldn't be used in this project specifically.
That being said, it is also not used in approver-policy (the only other library that uses makefile-modules today), so we actually don't really need it.

make/_shared/cert-manager/01_mod.mk
inteon: good catch, I'll change this. This is part of https://github.com/cert-manager/makefile-modules, so the fix will also improve approver-policy.

deploy/charts/istio-csr/templates/deployment.yaml
inteon: Yes, I think the entrypoint changes.

make/_shared/cert-manager/01_mod.mk
inteon: @SgtCoDFish https://github.com/cert-manager/makefile-modules/pull/17

test/carotation/workloads/httpbin.yaml
inteon: https://github.com/cert-manager/istio-csr/pull/226

inteon: Will have to merge this PR first.

make/_shared/cert-manager/01_mod.mk
inteon: We can merge that PR after this PR (or before).

test/carotation/workloads/httpbin.yaml
SgtCoDFish: Fair enough!

SgtCoDFish (approved):
/lgtm
/approve

This is clearly reasonable! Thanks for this, I know it's tonnes of work!

PR #226, opened 2024-01-08
Use upstream images instead of custom built images

inteon:
replaces:
- `quay.io/joshvanl_jetstack/curl` with `quay.io/curl/curl:latest`
- `quay.io/joshvanl_jetstack/httpbin:latest` with `kennethreitz/httpbin:latest`

### DIFF
--- a/test/carotation/workloads/sleep.yaml
+++ b/test/carotation/workloads/sleep.yaml
@@ -61,7 +61,7 @@ spec:
       serviceAccountName: sleep
       containers:
       - name: sleep
-        image: quay.io/joshvanl_jetstack/curl
+        image: quay.io/curl/curl:latest
         command: ["/bin/sleep", "3650d"]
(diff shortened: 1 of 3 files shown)


inteon:
blocked by https://github.com/cert-manager/istio-csr/pull/221

test/carotation/workloads/sleep.yaml
```diff
@@ -61,7 +61,7 @@ spec:
       serviceAccountName: sleep
       containers:
       - name: sleep
-        image: quay.io/joshvanl_jetstack/curl
+        image: quay.io/curl/curl:latest
```
SgtCoDFish, line 64: note: These curl images on quay.io are official, see https://github.com/curl/curl-container

I was suspicious of them initially because we so rarely see images on quay.io except our own!

SgtCoDFish (approved):
/lgtm
/approve

Definitely worth it!

PR #229, opened 2024-01-08
Upload junit and coverage results to ARTIFACTS folder

inteon:
Output junit reports to a file in the ARTIFACTS folder, so the tests show as a beautiful list in prow.

Also uploads coverage profile for the unit tests.


SgtCoDFish (approved):
/lgtm
/approve

I'm not sure if I'd go as far as to say they're "beautiful", but I think the outputs do look better 😉

PR #237, opened 2024-01-15
Upgrade repository-base module

inteon:
Could not self-upgrade because this upgrade changes a GH action.


inteon:
self-lgtm'ng
/approve
/lgtm

PR #251, opened 2024-01-24
feat: add the ability to specify certificate usages

fernandezcuesta:
Replaces #216

### DIFF
--- a/cmd/app/options/options.go
+++ b/cmd/app/options/options.go
@@ -28,6 +28,7 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	"k8s.io/client-go/rest"
 	cliflag "k8s.io/component-base/cli/flag"
+	cmapi "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
 	"k8s.io/klog/v2"
 	"k8s.io/klog/v2/klogr"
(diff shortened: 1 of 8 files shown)


kian:
@fernandezcuesta did you want to continue with this or have you addressed your own issue separately?  I ran into the same issue with AWS PCA + RAM and can open a rebased PR based on your diff, unless you plan to continue with this

SgtCoDFish:
Hey - this obviously should've been reviewed sooner. Sorry about that.

Given there are a lot of changes required here I'd suggest a new PR might be the way to go and I'm going to close this one. @kian maybe you're still interested?

A new PR would obviously need to address merge conflicts, but also wouldn't update the version in the chart since we handle that differently now.

Top tip would be to put a followup PR in [`#cert-manager-dev`](https://cert-manager.io/docs/contributing/#slack) in k8s slack, which would help to get a review.

Sorry again this took a long time to get eyes on!

hawksight:
Had a quick look through this and the chart changes look decent to me.
I'm unsure on the code side and whether this may still be necessary since all of @SgtCoDFish's hard work on making configuration dynamic.

I'm assuming here that AWS RAM enforces you to set the correct certificateUsages, hence adding the ability to set them here.

PR #262, opened 2024-01-30
add ability to provide additional DNS names to istiod certificate

apsega:
Add ability to provide additional istiod DNS certificates. Useful when istiod should be reachable from outside the Kubernetes cluster.

Signed-off-by: Edgaras <edgaras@apsega.lt>


JoshVanL:
Thanks! :) 
/lgtm
/approve
/ok-to-test

PR #267, opened 2024-02-08
Remove README header since it is already included in the artifacthub sidebar

inteon:
similar to https://github.com/cert-manager/approver-policy/pull/367


SgtCoDFish (approved):
/lgtm
/approve

I like this with the link to artifacthub!

PR #268, opened 2024-02-08
Fix values.yaml

inteon:
- The nodePort option was not documented in the values.yaml file
- configmapNamespaceSelector is a string, but was set to null by default

Prepares the chart for https://github.com/cert-manager/istio-csr/pull/258


deploy/charts/istio-csr/values.yaml
```diff
@@ -26,6 +26,9 @@ service:
   type: ClusterIP
   # Service port to expose istio-csr gRPC service.
   port: 443
+  # Service nodePort to expose istio-csr gRPC service.
+  # +docs:property
```
ThatsMrTalbot, line 30: Does nodeport need a type also?

PR #271, opened 2024-02-09
Bump to latest Istio version for E2E tests

MattiasGees:
This PR bumps the E2E tests to the latest Istio version. I also included an example of the IstioOperator config for 1.19, as that is also still an officially supported Istio version.


MattiasGees:
> Thanks for doing this, looking good, but I had actually merged this change before 1.16 release in Istio [istio.io/latest/news/releases/1.16.x/announcing-1.16/change-notes/#security](https://istio.io/latest/news/releases/1.16.x/announcing-1.16/change-notes/#security)
> 
> So we can update 1.16 and 1.7 example as well to leverage the simpler installation!

Done, updated the older versions as well.

SpectralHiss:
Thanks for the contribution!
I (soft) approve this as project reviewer.

please take a look at the corresponding website PR to update here:  https://github.com/cert-manager/website/pull/1424/files

inteon:
After merging this, we need to create a PR here too: https://github.com/cert-manager/testing/tree/master/config/jobs/cert-manager/istio-csr

inteon:
Seems like we should first change https://github.com/cert-manager/testing/tree/master/config/jobs/cert-manager/istio-csr, because the old tests stop working due to this change.
Better would be to fully move the set of istio versions to this repo (instead of the testing repo).

SpectralHiss:
> Seems like we should first change https://github.com/cert-manager/testing/tree/master/config/jobs/cert-manager/istio-csr, because the old tests stop working due to this change. Better would be to fully move the set of istio versions to this repo (instead of the testing repo).

I agree about the long-term update , currently just doing this:
https://github.com/cert-manager/testing/pull/962

make/config/istio/istio-config-1.16.2.yaml
```diff
@@ -1,57 +0,0 @@
...
-      # Change certificate provider to cert-manager istio agent for istio agent
-      caAddress: cert-manager-istio-csr.cert-manager.svc:443
-  components:
-    pilot:
-      k8s:
-        env:
-          # Disable istiod CA Sever functionality
-        - name: ENABLE_CA_SERVER
-          value: "false"
-        overlays:
```
inteon, line 22: @MattiasGees Do we no longer need these overlays?

make/config/istio/istio-config-1.16.7.yaml
```diff
@@ -0,0 +1,20 @@
...
+  namespace: istio-system
+spec:
+  profile: "minimal"
+  meshConfig:
+    # Change the following line to configure the trust domain of the Istio cluster.
+    trustDomain: foo.bar
+  values:
+    global:
+      # Change certificate provider to cert-manager istio agent for istio agent
+      caAddress: cert-manager-istio-csr.istio-system.svc:443
```
inteon, line 13:
```suggestion
      caAddress: cert-manager-istio-csr.cert-manager.svc:443
```
The e2e setup installs istio-csr in the cert-manager namespace.

inteon: I fixed this in https://github.com/cert-manager/istio-csr/pull/271/commits/5ba84bd77bdec205e0e9da0f6747f5bf867ae8aa.

make/config/istio/istio-config-1.16.2.yaml
inteon: I just found this: https://github.com/cert-manager/website/pull/1424

SpectralHiss: yeah I had fixed it in Istio 1.16 but never found the time to update it until @MattiasGees reminded me!

PR #274, opened 2024-02-16
Hef/csr p384 generation fix

### LINKED ISSUE #246: CSR generation always defaults to P256 curve due to missing parameter
We are currently using istio-csr in combination with a cert-manager Vault issuer. In Vault we configured the pki-roles to only allow ECDSA keys with 384-bit curves. As soons as istio-csr tries to sign CSRs using the Vault issuer it fails:

> 2024-01-22T10:53:30.030036Z    error    klog    failed to fetch initial serving certificate: failed to sign serving certificate: failed to wait for CertificateRequest istio-system/istio-csr-s7wf8 to be signed: created CertificateRequest has failed: [{Approved True
2024-01-22 10:53:29 +0000 UTC cert-manager.io Certificate request has been approved by cert-manager.io} {Ready False
2024-01-22 10:53:29 +0000 UTC Failed Vault failed to sign certificate: failed to sign certificate by vault: Error making API request. URL: POST https://xxxxxxxxxxxxxxxx/v1/pki_int/sign/xxxxxxxxxx 
Code: 400. Errors:  * role requires a minimum of a 384-bit key, but CSR's key is 256 bits}]: error running cert-manager-istio-csr

Quickly glancing at the code it seems that the `GenCSR` method of istios `pkiutil` 'expects' a parameter called `ECCCurve` and defaults to a 256-bit curve in case its not provided: https://github.com/istio/istio/blob/master/security/pkg/pki/util/generate_csr.go#L42 

Unfortunately in the `fetchCertificate` method the `opts` map lacks this option, it is neither derived from the `ServingCertificateKeySize` nor is it settable through a dedicated parameter: https://github.com/cert-manager/istio-csr/blob/e82df7e50b460e86668b78737a5db3a29891f13c/pkg/tls/tls.go#L288

Any chance this could implemented to allow the usage of ECDSA keys with a curve size larger than 256 bits?

SpectralHiss:
Fixes #246 

Add an e2e test for ECC that needs adding in "testing" jobs, it can serve as an example

### DIFF
--- a/deploy/charts/istio-csr/README.md
+++ b/deploy/charts/istio-csr/README.md
@@ -264,7 +264,7 @@ Create the default certificate as part of install.
 > 2048
 > ```
 
-Number of bits to use for istiod-tls RSAKey
+Number of bits to use for istiod-tls Key
 #### **app.server.clusterID** ~ `string`
 > Default value:
 > ```yaml
@@ -301,7 +301,7 @@ Container port to serve istio-csr gRPC service.
 > 2048
 > ```
 
-Number of bits to use for the server's serving certificate (RSAKeySize).
+Number of bits to use for the server's serving certificate, can only be 256 or 384 when signature algorithm is ECDSA.
 #### **app.server.serving.signatureAlgorithm** ~ `string`
 > Default value:
 > ```yaml
--- a/make/test-ecc.mk
+++ b/make/test-ecc.mk
@@ -1,0 +1,32 @@
+# Copyright 2023 The cert-manager Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+.PHONY: test-ecc
+## ecc test
+## @category Testing
+test-ecc: kind_cluster_name := "istio-csr-ecc"
+test-ecc: e2e-setup-cert-manager oci-load-manager | $(bin_dir)/scratch/istioctl-$(ISTIO_VERSION) $(NEEDS_KUBECTL) $(NEEDS_HELM) $(NEEDS_KIND) $(NEEDS_GOJQ)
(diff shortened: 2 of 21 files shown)


make/test-ecc.mk
```diff
@@ -0,0 +1,32 @@
...
+.PHONY: test-ecc
+## ecc test
+## @category Testing
+test-ecc: kind_cluster_name := "istio-csr-ecc"
+test-ecc: e2e-setup-cert-manager oci-load-manager | $(bin_dir)/scratch/istioctl-$(ISTIO_VERSION) $(NEEDS_KUBECTL) $(NEEDS_HELM) $(NEEDS_KIND) $(NEEDS_GOJQ)
+	$(eval oci_image_tar := $(bin_dir)/scratch/image/oci-layout-manager.$(oci_manager_image_tag).docker.tar)
+
+	ARTIFACTS=$(ARTIFACTS) \
+	ISTIO_CSR_IMAGE=$(oci_manager_image_name_development) \
+	ISTIO_CSR_IMAGE_TAR=$(oci_image_tar) \
```
inteon, line 24: It looks like this variable is no longer used in the carotation test, I think it can be removed here too?

deploy/charts/istio-csr/README.md
```diff
@@ -267,7 +267,7 @@ Create the default certificate as part of install.
 > 2048
 > ```
 
-Number of bits to use for istiod-tls RSAKey
+Number of bits to use for istiod-tls Key
```
SgtCoDFish, line 270: comment: as mentioned in standup, this is the wrong abstraction for ECDSA but this kind of thing is used in istio and cert-manager so it's totally fine here!

SgtCoDFish (commented):
I'd be happy to approve this I think! In standup Houssem said he'd look into getting the ECDSA tests running in prow, and once they're running and passing I think this is good to go from my view. We can always review fixes if needed!

make/test-ecc.mk
SpectralHiss: sorry where do you see that it's no longer used?

deploy/charts/istio-csr/README.md
SpectralHiss: I lied a little bit and it turns out they do use: "ECC_CURVE" : "P384" or 256 currently in Istio.
However it is true that the only supported ecc curves will be those two for the foreseable future!
If you want I can add a new key for ecc_curve and change it, it's not biggie

SpectralHiss:
> I'd be happy to approve this I think! In standup Houssem said he'd look into getting the ECDSA tests running in prow, and once they're running and passing I think this is good to go from my view. We can always review fixes if needed!

This is the PR adding the prow jobs
https://github.com/cert-manager/testing/pull/969

inteon (commented):
/approve
/lgtm
Thanks for the PR!

PR #292, opened 2024-03-27
Fix linters

inteon:
Reduces the list of ignored linters by fixing the go code.


SgtCoDFish (approved):
/lgtm
/approve

Thanks!

PR #298, opened 2024-04-22
Upgrade repository-base module

inteon:
This cannot be upgraded by the self-upgrade script since it is changing the GH actions.


inteon:
self-lgtm'ng since this is already merged in makfiles-modules repo

PR #306, opened 2024-04-26
Fix gosec linter issues and enable linter

inteon:
Checked all found issues and applied fixes/ #nosec statements.
Now that all issues are fixed, I removed the gosec linter from the ignored section so future issues will be detected.

### DIFF
--- a/pkg/tls/tls.go
+++ b/pkg/tls/tls.go
@@ -241,6 +241,7 @@ func (p *Provider) Config(ctx context.Context) (*tls.Config, error) {
 
 		if conf != nil {
 			return &tls.Config{
+				MinVersion:         tls.VersionTLS12,
 				GetConfigForClient: p.getConfigForClient,
 				ClientAuth:         tls.RequireAndVerifyClientCert,
 			}, nil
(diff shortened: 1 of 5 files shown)


pkg/tls/tls.go
```diff
@@ -241,6 +241,7 @@ func (p *Provider) Config(ctx context.Context) (*tls.Config, error) {
 
 		if conf != nil {
 			return &tls.Config{
+				MinVersion:         tls.VersionTLS12,
 				GetConfigForClient: p.getConfigForClient,
 				ClientAuth:         tls.RequireAndVerifyClientCert,
 			}, nil
```
SgtCoDFish, line 247: comment: surprised this was recommended!

```console
$ go doc tls Config.MinVersion
package tls // import "crypto/tls"

type Config struct {
    // MinVersion contains the minimum TLS version that is acceptable.
    // 
    // By default, TLS 1.2 is currently used as the minimum. TLS 1.0 is the minimum
    // supported by this package.
    // 
    // The server-side default can be reverted to TLS 1.0 by including the value
    // "tls10server=1" in the GODEBUG environment variable.
    MinVersion uint16

    // ... other fields elided ...
}
```

I mean, it doesn't hurt... but it looks like a no-op?

SgtCoDFish (approved):
/lgtm
/approve

Thanks 🚀

PR #315, opened 2024-05-13
Helm: set linux nodeSelector by default

inteon:
Similar to cert-manager and csi-driver-spiffe, set the nodeSelector to kubernetes.io/os: linux by default.


SgtCoDFish (approved):
/lgtm
/approve

Thanks!

PR #343, opened 2024-07-15
Add ability to configure issuer at runtime

SgtCoDFish:
This commit allows users to specify the name of a ConfigMap which should be watched for issuer configuration. Importantly, this allows for istio-csr to be installed at the same time as cert-manager without needing an issuer to already exist

The helm chart assumes that the ConfigMap is in the same namespace as the istio-csr Deployment.

This is very similar to https://github.com/cert-manager/csi-driver-spiffe/pull/141 which implemented similar functionality for csi-driver-spiffe.

While this is similar to csi-driver-spiffe, a key difference is that the runtime configuration configmap watcher in this PR is a controller-runtime Runnable.

### DIFF
--- a/make/config/istio-csr-values.yaml
+++ b/make/config/istio-csr-values.yaml
@@ -1,4 +1,4 @@
-replicaCount: 3
+replicaCount: 2
 
 service:
   port: 443
--- a/pkg/certmanager/certmanager.go
+++ b/pkg/certmanager/certmanager.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync"
 	"time"
 
 	apiutil "github.com/cert-manager/cert-manager/pkg/api/util"
@@ -28,10 +29,13 @@ import (
 	cmversioned "github.com/cert-manager/cert-manager/pkg/client/clientset/versioned"
 	cmclient "github.com/cert-manager/cert-manager/pkg/client/clientset/versioned/typed/certmanager/v1"
 	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	ctrlmgr "sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
 const (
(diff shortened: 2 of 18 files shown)


pkg/certmanager/certmanager.go
```diff
@@ -211,3 +274,187 @@ func (m *manager) waitForCertificateRequest(ctx context.Context, log logr.Logger
...
+	if !exists || len(issuerRef.Group) == 0 {
+		dataErrs = append(dataErrs, fmt.Errorf("missing key/value in ConfigMap data; %s", issuerGroupKey))
+	}
+
+	if len(dataErrs) > 0 {
+		return errors.Join(dataErrs...)
+	}
+
+	// we now have a full issuerRef
+	// TODO: check if the issuer exists by querying for the CRD?
```
SgtCoDFish, line 318: note: This TODO is copied from csi-driver-spiffe - this feature merged in csi-driver-spiffe without this check.

make/config/istio-csr-values.yaml
```diff
@@ -1,4 +1,4 @@
-replicaCount: 3
```
SgtCoDFish, line 1: note: this is technically an unrelated change. I reduced the count by 1 here because istio-csr is leader-elected and the spare pods aren't really used in the e2e tests we do at all.

I think there's some value in having a spare running, but I couldn't see a reason to have _two_ spares running!

pkg/certmanager/certmanager.go
```diff
@@ -79,20 +108,54 @@ type Bundle struct {
...
-		client: client.CertmanagerV1().CertificateRequests(opts.Namespace),
-		opts:   opts,
+		log: log.WithName("cert-manager"),
+
+		kubernetesClient:  k8sClient,
+		certManagerClient: cmClient.CertmanagerV1().CertificateRequests(opts.Namespace),
+		opts:              opts,
+
+		activeIssuerRef:      activeIssuerRef,
+		activeIssuerRefMutex: &sync.RWMutex{},
```
ThatsMrTalbot, line 144: NIT: We already only use a pointer to `manager` so the RWMutex does not need to be a pointer, it can be a concrete type and not need initialisation.

SgtCoDFish: Yeah that's totally fair. I should've changed that when I had to initialise it in the unit tests. I'll change 👍

SgtCoDFish:
I've added a second commit with the suggested change and a minor extra one (to remove a pointless deployment running in the e2e test)

SgtCoDFish:
The merging of https://github.com/cert-manager/testing/pull/1044 made the test output go a bit crazy.

I reworked the commits. My change to use a single service for the runtime config e2e test revealed a race condition in how we wait for pods to be ready. I've addressed that and fixed it for the mtls e2e test too.

SgtCoDFish (commented):
/approve

PR #345, opened 2024-07-16
Add updated versions of istio config to enable tests

SgtCoDFish:
This will allow us to test the latest patch release of existing versions and add support for testing 1.21.x and 1.22.x

I tested all of these configs locally and the e2e tests passed for all.


inteon (commented):
/approve
/lgtm

PR #348, opened 2024-07-17
Increase timeout for namespace e2e test polling

SgtCoDFish:
The `CA Root Controller all namespaces should have valid configs in` test regularly flakes, causing CI to fail.

As a first step, try to reduce flakes by increasing this timeout, on the assumption that the flakes are caused by a heavy load in CI or some other thing which causes them to time out.

Also:

- Changes to use variables when calling the `Poll` function, which makes the arguments clearer to reason about.
- Adds `no-color` to ginkgo when running in CI to make output more readable

### DIFF
--- a/make/test-e2e.mk
+++ b/make/test-e2e.mk
@@ -73,6 +73,15 @@ test-e2e-deps: e2e-create-cert-manager-istio-resources
 test-e2e-deps: install
 test-e2e-deps: e2e-setup-istio
 
+CI ?=
+EXTRA_GINKGO_FLAGS :=
+
+# In Prow, the CI environment variable is set to "true"
+# See https://docs.prow.k8s.io/docs/jobs/#job-environment-variables
+ifeq ($(CI),true)
+EXTRA_GINKGO_FLAGS += --no-color
+endif
+
 .PHONY: test-e2e
 ## e2e end-to-end tests
 ## @category Testing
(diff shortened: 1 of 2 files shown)


SgtCoDFish:
Interesting results from the test runs on this PR:

[This job](https://prow.infra.cert-manager.io/view/gs/cert-manager-prow-artifacts/pr-logs/pull/cert-manager_istio-csr/348/pull-cert-manager-istio-csr-istio-v1-18/1813538947716878336) for v1.18 took 44s and would have flaked before this PR

All other instances of the flaky test passed in a few seconds!

SgtCoDFish:
On the second run when I pushed more commits, the ecc test flaked for an unrelated reason.

As for e2e tests:

- [v1.20](https://prow.infra.cert-manager.io/view/gs/cert-manager-prow-artifacts/pr-logs/pull/cert-manager_istio-csr/348/pull-cert-manager-istio-csr-istio-v1-20/1813547872642142208) would have flaked before this change. It took 44s.
- [v1.21](https://prow.infra.cert-manager.io/view/gs/cert-manager-prow-artifacts/pr-logs/pull/cert-manager_istio-csr/348/pull-cert-manager-istio-csr-istio-v1-21/1813547872717639680) would have flaked too. Also took 44s.

Other e2e tests all had the flaky test pass in a couple of seconds.

The second run included the change to set `--no-color` in Ginkgo for e2e tests on prow, and that worked as expected.

SgtCoDFish:
On the third run (after I removed the commit which broke the ecc test) I saw the same story:

- [v1.18](https://prow.infra.cert-manager.io/view/gs/cert-manager-prow-artifacts/pr-logs/pull/cert-manager_istio-csr/348/pull-cert-manager-istio-csr-istio-v1-18/1813561151280123904) took 44s and would have flaked before this PR
- [v1.19](https://prow.infra.cert-manager.io/view/gs/cert-manager-prow-artifacts/pr-logs/pull/cert-manager_istio-csr/348/pull-cert-manager-istio-csr-istio-v1-19/1813561151443701760) would've flaked too, also 44s
- All other versions took seconds.

inteon (commented):
Hopefully this fixes the flake. However, let's stay alert for any weirdness we find in the testing logic that might explain these kind of flakes.
/approve
/lgtm

PR #354, opened 2024-07-18
add flag to set log-format

seankhliao:
I would like logs to be output in json format.

This uses the same logs api libraries as cert-manager itself,
though it doesn't expose the full logging configuration surface,
just an extra `--log-format` flag.

Example original output:

```
I0718 21:24:43.654238   65256 options.go:113] "------------------------------------------------------------------------------------------------------------"
I0718 21:24:43.654292   65256 options.go:114] "WARNING!: --root-ca-file is not defined which means the root CA will be discovered by the configured issuer."
I0718 21:24:43.654298   65256 options.go:115] "WARNING!: It is strongly recommended that a root CA bundle be statically defined."
I0718 21:24:43.654303   65256 options.go:116] "------------------------------------------------------------------------------------------------------------"
I0718 21:24:43.657170   65256 server.go:208] "Starting metrics server" logger="controller-runtime.metrics"
I0718 21:24:43.657222   65256 server.go:247] "Serving metrics server" logger="controller-runtime.metrics" bindAddress="0.0.0.0:9402" secure=false
I0718 21:24:43.657319   65256 server.go:83] "starting server" name="health probe" addr="[::]:6060"
I0718 21:24:43.657582   65256 leaderelection.go:250] attempting to acquire leader lease istio-system/istio-csr...
I0718 21:24:43.933090   65256 internal.go:525] "Stopping and waiting for non leader election runnables" logger="manager"
I0718 21:24:43.933159   65256 internal.go:529] "Stopping and waiting for leader election runnables" logger="manager"
I0718 21:24:43.933176   65256 internal.go:537] "Stopping and waiting for caches" logger="manager"
I0718 21:24:43.933216   65256 internal.go:541] "Stopping and waiting for webhooks" logger="manager"
I0718 21:24:43.933239   65256 internal.go:544] "Stopping and waiting for HTTP servers" logger="manager"
I0718 21:24:43.933283   65256 server.go:68] "shutting down server" name="health probe" addr="[::]:6060"
E0718 21:24:43.933364   65256 internal.go:499] "error received after stop sequence was engaged" err="context canceled" logger="manager"
I0718 21:24:43.933399   65256 server.go:254] "Shutting down metrics server with timeout of 1 minute" logger="controller-runtime.metrics"
I0718 21:24:43.933575   65256 internal.go:548] "Wait completed, proceeding to shutdown the manager" logger="manager"
[...]

PR #356, opened 2024-07-22
Test with JSON logs, improve logs aesthetics for JSON

SgtCoDFish:
The big warning messages don't look great in JSON logs because they're multiline. This consolidates those messages into one line.

It also slightly tweaks the parsing of the log format to ensure we produce a good log message.

Also uses JSON logging in e2e tests.


SgtCoDFish:
This is a followup to  #354

PR #357, opened 2024-07-23
Implement client cert auth

### LINKED ISSUE #279: Istio sidecar can only request new cert using istio-token
When using Citadel CA istio sidecars can fetch new certificates using the previous cert while it's valid even when `istio-token` is expired.

When using istio-csr / cert-manager / Vault new certs can only be fetched while `istio-token` is valid but old cert appears to be ignored. Or at least it only fetches new certs with a valid `istio-token`.

Is this behavior expected or not? I've expected for istio sidecar to renew cert with the previous one.

kubernetes version: v1.26.7
istio version: v1.20.2
cert-manager version: v1.12.7
istio-csr version: v0.8.1
vault version: v1.15.2

wtzhang23:
## Description
* For VMs, rotation of k8s service account token (istio-token) is not reliable
* Reuse logic in Istiod which uses the peer cert to identify the workload on refresh
* Fixes #279


SgtCoDFish:
Hey, thanks for raising this! I think this makes sense and it's a well made PR!

I don't really know how to evaluate the security impact of this (I'm not an istio expert). I don't know if having client cert auth enabled by default is a safe thing in all cases. It seems fairly safe in this context, though, since obviously the admin will have chosen which certs go into the trust store to use for checking client certs.

I'm reluctant to enable this by default just because it expands the auth surface area and as said above I'm nervous of that.

Do you think we could add a boolean flag (and a corresponding Helm chart bool) which allows users to opt-in to client cert auth? I feel like then it's super easy to merge this feature.

wtzhang23:
Added the flag in a separate `authenticators` block in case there comes a need for more authenticators in the future. Istio seems to support based on the highlighted [code](https://github.com/istio/istio/blob/afdad15000541f1a4c82f58918868511553e1a87/pilot/pkg/bootstrap/server.go#L356-L384):
* client cert
* k8s sa token
* jwt
* cidrs
as well as toggles that disable authentication in a trusted network (my guess is they only need the cert for multicluster / SPIFFE federation).

SgtCoDFish (approved):
/lgtm
/approve

Thank you for this! Really appreciate the work you've put into it!

With this being gated behind a flag I think this is a really safe bet to add. I'm hoping to do an istio-csr release pretty soon, so please do test it when the release is out 🚀

PR #364, opened 2024-07-31
Add retries for issuing initial serving cert

SgtCoDFish:
This enables istio-csr to be installed concurrently with cert-manager and to continue trying to operate until the initial serving cert is available

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/fsnotify/fsnotify v1.7.0
 	github.com/go-logr/logr v1.4.2
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
+	github.com/lestrrat-go/backoff/v2 v2.0.8
 	github.com/onsi/ginkgo/v2 v2.19.0
 	github.com/onsi/gomega v1.33.1
 	github.com/prometheus/client_golang v1.19.1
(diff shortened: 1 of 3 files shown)


wallrj (approved):
I haven't tested this, but it looks like it should allow the istio-csr process to:
1. first wait for a ConfigMap to exist
2. wait for the ConfigMap to contain all of: issuer-name, issuer-kind, and issuer-group 
3. create cert-manager CertificateRequest resources (with the given issuer-ref) until one of them gets signed.
4. Give up and exit with an error if this process is performed more than 250 times.


I don't really understand why the serving certificate for the  istio-csr gRPC service has to use the same cert-manager issuer is used for the CRs it generates for the istio workload certificates. In fact I would expect the serving certificate for the gRPC server to be signed by a different (platform) CA than the mTLS certificates of the workloads running **on** the platform.

I'd have thought that the istio-csr Pod could just mount a Secret containing the serving certificate and have that Secret populated by a cert-manager Certificate, as and when cert-manager gets installed. 
I suppose this would cause the Istio-csr helm deployment with `--wait` to fail...but nevertheless that would be a simple and easily understandable process for installing istio-csr in parallel with cert-manager. 

I just don't know enough about how istio-csr works. Don't want to hold you up though so I'll add an lgtm and we can do thorough testing when you release the alpha.

/approve
/lgtm
/hold incase there's anything else to be changed.

PR #369, opened 2024-08-02
Add security context to istio-csr deployment

### LINKED ISSUE #367: Configure security context of the istio-csr deployment
## Description
* Our Kubernetes cluster requires deployments to have as minimal of a security context as possible, including infrastructure components
* This issue proposes we set the privileges of the Deployment template to a minimal one like [Istiod](https://github.com/istio/istio/blob/master/manifests/charts/istio-control/istio-discovery/templates/deployment.yaml#L195-L205)

wtzhang23:
## Description
* The default security context may be too permissive compared to what a cluster operator requires; configure it to be as restrictive as possible.
* Allow configuring the seccompprofile in case more restrictive custom policies need to be applied
* Fixes #367

### DIFF
--- a/deploy/charts/istio-csr/templates/deployment.yaml
+++ b/deploy/charts/istio-csr/templates/deployment.yaml
@@ -121,16 +121,7 @@ spec:
           {{- toYaml .Values.resources | nindent 12 }}
 
         securityContext:
-          allowPrivilegeEscalation: false
-          readOnlyRootFilesystem: true
-          runAsNonRoot: true
-          capabilities:
-            drop:
-            - ALL
-          {{- if .Values.seccompProfile }}
-          seccompProfile:
-            {{ toYaml .Values.seccompProfile | indent 14 }}
-          {{- end }}
+          {{- toYaml .Values.securityContext | nindent 12 }}
 
       {{- if .Values.volumes }}
       volumes:
--- a/deploy/charts/istio-csr/templates/deployment.yaml
+++ b/deploy/charts/istio-csr/templates/deployment.yaml
@@ -120,6 +120,18 @@ spec:
         resources:
           {{- toYaml .Values.resources | nindent 12 }}
 
+        securityContext:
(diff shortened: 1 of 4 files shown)


hawksight:
@SgtCoDFish - [flakey test](https://storage.googleapis.com/cert-manager-prow-artifacts/pr-logs/pull/cert-manager_istio-csr/369/pull-cert-manager-istio-csr-test-ecc/1819299683034468352/build-log.txt) and worth a retry?

Or something obviously wrong I am missing?

deploy/charts/istio-csr/templates/deployment.yaml
```diff
@@ -120,6 +120,18 @@ spec:
         resources:
           {{- toYaml .Values.resources | nindent 12 }}
 
+        securityContext:
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
+          runAsNonRoot: true
+          capabilities:
+            drop:
+            - ALL
```
SgtCoDFish, line 129: suggestion: I think it's a good idea to enable the security context and set sane defaults - clearly these are pretty good defaults since the e2e tests passed without an issue.

That said, I think setting a restrictive securityContext like this - with no option to disable it - might cause issues for some users.

Can we add the securityContext to the values.yaml? I think it seems fine to have these defaults in values.yaml, but at least that will allow users to opt-out!

SgtCoDFish (changes requested):
Thanks for this; we discussed it a bit in our standup this morning and ended up with a suggestion. What do you think?

deploy/charts/istio-csr/templates/deployment.yaml
wtzhang23: Yep, kinda autopiloted copying Istio's helm charts but this makes more sense

SgtCoDFish (approved):
/lgtm
/approve

Thanks for this 😁

PR #378, opened 2024-08-14
feat: add ability to specify extra objects to apply along with the chart

ThatsMrTalbot:
This is useful because istio-csr depends on an external issuer to function. The ability to configure that issuer when deploying is desirable.

### DIFF
--- a/deploy/charts/istio-csr/values.yaml
+++ b/deploy/charts/istio-csr/values.yaml
@@ -325,3 +325,15 @@ nodeSelector:
 
 # Labels to apply to all resources
 commonLabels: {}
+
+# Create resources alongside installing istio-csr, via Helm values. Can accept an array of YAML-formatted
+# resources. Each array entry can include multiple YAML documents, separated by '---'
+#
+# For example:
+# extraObjects:
+#   - |
+#     apiVersion: v1
+#     kind: ConfigMap
+#     metadata:
+#       name: '{{ template "cert-manager-istio-csr.fullname" . }}-extra-configmap'
+extraObjects: []
--- a/deploy/charts/istio-csr/README.md
+++ b/deploy/charts/istio-csr/README.md
@@ -548,5 +548,23 @@ Kubernetes node selector: node labels for pod assignment.
 > ```
 
 Labels to apply to all resources
+#### **extraObjects** ~ `array`
+> Default value:
+> ```yaml
+> []
+> ```
+
+Create resources alongside installing istio-csr, via Helm values. Can accept an array of YAML-formatted resources. Each array entry can include multiple YAML documents, separated by '---'  
+  
+For example:
+
+```yaml
+extraObjects:
+  - |
+    apiVersion: v1
+    kind: ConfigMap
+    metadata:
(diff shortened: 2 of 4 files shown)


deploy/charts/istio-csr/values.yaml
```diff
@@ -298,3 +298,14 @@ nodeSelector:
...
+
+# Create dynamic manifests via values.
+#
+# For example:
+# extraObjects:
+#   - |
+#     apiVersion: v1
+#     kind: ConfigMap
+#     metadata:
+#       name: '{{ template "cert-manager.fullname" . }}-extra-configmap'
```
SgtCoDFish, line 310: suggestion: This example doesn't actually work locally for me because `cert-manager.fullname` doesn't seem to be defined. It's minor, but I think the example would be better without the template bit

```suggestion
# For example:
# extraObjects:
#   - |
#     apiVersion: v1
#     kind: ConfigMap
#     metadata:
#       name: 'some-extra-configmap'
```

deploy/charts/istio-csr/values.yaml
```diff
@@ -298,3 +298,14 @@ nodeSelector:
 
 # Labels to apply to all resources
 commonLabels: {}
+
+# Create dynamic manifests via values.
```
SgtCoDFish, line 302: suggestion: Rewording and a bit of extra usage info:

```suggestion
# Create resources alongside installing istio-csr, via Helm values. Can accept an array of YAML-formatted
# resources. Each array entry can include multiple YAML documents, separated by '---'
```

SgtCoDFish (changes requested):
This looks good and testing locally it works. I used the following values.yaml:

```yaml
extraObjects:
  - |
    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: 'some-extra-configmap'
  - |
    apiVersion: v1
    kind: Namespace
    metadata:
      name: sandbox
    ---
    apiVersion: cert-manager.io/v1
    kind: Issuer
    metadata:
      name: selfsigned-issuer
      namespace: sandbox
    spec:
      selfSigned: {}
    ---
    apiVersion: cert-manager.io/v1
    kind: Certificate
    metadata:
      name: root-ca
      namespace: sandbox
    spec:
      isCA: true
      commonName: root-ca
      secretName: root-ca
      privateKey:
        algorithm: ECDSA
        size: 256
      issuerRef:
        name: selfsigned-issuer
        kind: Issuer
        group: cert-manager.io
    ---
    apiVersion: cert-manager.io/v1
    kind: Issuer
    metadata:
      name: root-ca-issuer
      namespace: sandbox
    spec:
      ca:
        secretName: root-ca
```

And that correctly created all resources. I added some suggestions on rewording the doc comment but you might want to reword - I'm easy really.

What do you think?

deploy/charts/istio-csr/values.yaml
ThatsMrTalbot: Copy paste error, should be "cert-manager-istio-csr.fullname"

SgtCoDFish (approved):
/lgtm
/approve

Cheers 🚀

PR #392, opened 2024-08-28
feat: make readiness probe optional

ThatsMrTalbot:


SgtCoDFish:
> I originally was planning to alter the checks if the issuer was not set, however we default the issuer in the Helm config. This means a user would have to explicitly unset three Helm values to get into this mode.

I think this is a bit of a tangent to the issue at hand, no? Like, it's not a great UX and I rather that the issuer didn't have a default value but we're kinda forced to accept it now for backwards compat reasons.

Although I guess it's a better UX to have `app.certmanager.issuer.enabled` and just have to set it to false than having to do the blanking. I think I'd support that!

PR #395, opened 2024-09-02
feat: allow "runtime-only" configuration without default issuer

ThatsMrTalbot:
### Helm

- Added `app.certmanager.issuer.enabled` to disable "default" non runtime issuer.

### app.go

- Wrapped the health checks to only run when issuer config is available. This means that we still go "not ready" if invalid issuer config is provided or the runtime config obtained is invalid. 

### options.go

- Added `--issuer-enabled` flag to disable "default" non runtime issuer. By setting this to false you can start istio-csr up and have it idle waiting for runtime config (while being ready).

### certmanager.go

- Updated the runtime config watcher to prune subscriptions that are not in use. Functions like `WaitForIssuerConfig` only need a single channel event then stopped monitoring it. The addition of the ability to "close" the subscription means these functions can "unsubscribe".

- Moved `WaitForIssuerConfig` into the manager, it makes sense in my mind for it to be there so multiple things can call it if needed.

### configmap.go / tls.go

- The original implementation of `tls.Interface.RootCAs` was blocking, this meant that if you had no "default" issuer config and the runtime config was not yet configured then the configmap controller Reconcile would block forever and delay graceful shutdown. The new implementation of `RootCAs` accepts a context so the controller can shut down gracefully.

### DIFF
--- a/pkg/certmanager/certmanager.go
+++ b/pkg/certmanager/certmanager.go
@@ -50,6 +50,9 @@ type Options struct {
 	// Namespace is the namespace that CertificateRequests will be created in.
 	Namespace string
 
+	// DefaultIssuerEnabled indicates the default issuer is enabled
+	DefaultIssuerEnabled bool
+
 	// IssuerRef is used as the issuerRef on created CertificateRequests.
 	IssuerRef cmmeta.ObjectReference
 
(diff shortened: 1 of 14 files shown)


pkg/certmanager/certmanager.go
```diff
@@ -75,12 +78,45 @@ type Signer interface {
...
+type IssuerChangeSubscription struct {
+	C <-chan *cmmeta.ObjectReference
+
+	// The same channel as above is mirrored in c, but without the "<-"
+	// restriction. This allows the channel to be written to by this package.
+	c chan *cmmeta.ObjectReference
+
+	lock   sync.Mutex
+	closed bool
+}
```
SgtCoDFish, line 92: suggestion: I'm a bit squeamish about having two variables with the same name differing only in casing - I personally find that pretty hard to keep in my head.

I like `C` for the exported one, but maybe something like `sendCh`, `internalCh` or something else instead of `c`?

pkg/istiodcert/worker.go
```diff
@@ -57,7 +57,7 @@ type DynamicIstiodCertProvisioner struct {
 
 	issuerRefMutex sync.Mutex
 
-	issuerChangeChan <-chan *cmmeta.ObjectReference
+	issuerChangeChan *certmanager.IssuerChangeSubscription
```
SgtCoDFish, line 60: nitpick (non-blocking): maybe rename because this isn't a channel any more?

```suggestion
	issuerChangeSub *certmanager.IssuerChangeSubscription
```

SgtCoDFish (changes requested):
One little comment - what do you think? It's not absolutely required!

pkg/istiodcert/worker.go
ThatsMrTalbot: Makes sense - updated

pkg/certmanager/certmanager.go
ThatsMrTalbot: Changed it to `sendChannel`

SgtCoDFish (approved):
/lgtm
/approve

Cheers! 🚀

PR #397, opened 2024-09-02
fix: do not track "originalIssuerRef" if default issuer is disabled

ThatsMrTalbot:
This is a small fix that prevents originalIssuerRef being set if `--issuer-enabled` is false. This means `HasIssuerConfig` will return false as desired until a runtime issuer config is provided.

Without this fix the `--issuer-*` flags must also be empty on top if `--issuer-enabled=false`


SgtCoDFish (approved):
/lgtm
/approve

Thanks!

PR #422, opened 2024-10-08
fix: expose ALPN in TLS handshake

howardjohn:
New versions of gRPC-go are enforcing the `h2` ALPN to be presented during the TLS handshake. See
https://pkg.go.dev/google.golang.org/grpc/internal/envconfig#pkg-variables `GRPC_ENFORCE_ALPN_ENABLED`. The TLS server here isn't automatically getting this set due to usage of GetConfigForClient.

This properly sets it.

Without this, istio-csr will be incompatible with Istio 1.24, which upgrades the gRPC version. Note this can be worked around by setting `GRPC_ENFORCE_ALPN_ENABLED=false` on the proxy container, which Istio is able to do -- so there is an escape hatch for users.

The Istio logs look like
`"transport: authentication handshake failed: credentials: cannot check peer: missing selected ALPN property"`

### DIFF
--- a/pkg/tls/tls.go
+++ b/pkg/tls/tls.go
@@ -427,8 +427,11 @@ func (p *Provider) fetchCertificate(ctx context.Context) (time.Time, error) {
 	p.tlsConfig = &tls.Config{
 		MinVersion:   tls.VersionTLS12,
 		Certificates: []tls.Certificate{tlsCert},
-		ClientAuth:   tls.VerifyClientCertIfGiven,
-		ClientCAs:    peerCertVerifier.GetGeneralCertPool(),
+		// Advertise ALPN, required in modern gRPC versions
+		// Typically gRPC sets this for us, but since this tls.Config ultimately gets returned in GetConfigForClient it doesn't.
+		NextProtos: []string{"h2"},
+		ClientAuth: tls.VerifyClientCertIfGiven,
+		ClientCAs:  peerCertVerifier.GetGeneralCertPool(),
 		VerifyPeerCertificate: func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
 			err := peerCertVerifier.VerifyPeerCert(rawCerts, verifiedChains)
 			if err != nil {


howardjohn:
cc @inteon

inteon (commented):
Thank you for the bug report and fix.

Could you add a tests/ modify our tests to make sure GRPC_ENFORCE_ALPN_ENABLED now works & will keep working in the future?

howardjohn:
> Also, do you know where in code this is normally auto-set when we are not using GetConfigForClient?

`credentials.NewTLS()` from gRPC auto adds it

> Could you add a tests/ modify our tests to make sure GRPC_ENFORCE_ALPN_ENABLED now works & will keep working in the future?

Should we just add a test for Istio 1.24 directly? Can use a pre-release in the meantime

inteon:
Adding an Istio 1.24 test sounds like a great solution.

inteon:
@howardjohn In order to test istio 1.24, we will have to add a file here: https://github.com/cert-manager/istio-csr/tree/main/make/config/istio.

We can locally test the test using the following command:
`ISTIO_VERSION="1.24.0-alpha.0" make test-e2e`

howardjohn:
> @howardjohn In order to test istio 1.24, we will have to add a file here: https://github.com/cert-manager/istio-csr/tree/main/make/config/istio.
> 
> We can locally test the test using the following command: `ISTIO_VERSION="1.24.0-alpha.0" make test-e2e`

Thanks, added it and also https://github.com/cert-manager/testing/pull/1066/files

howardjohn:
The errors were in the istio-proxy logs in my experience

SgtCoDFish:
Adding even though this has merged: I saw these errors in istio-proxy in the e2e tests without this change:

```text
2024-10-25T10:13:04.992535Z     error   citadelclient   failed to sign CSR: create certificate: rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: credentials: cannot check peer: missing selected ALPN property"
```

I

PR #429, opened 2024-10-25
Use a default (but configurable) test file for istio-csr

SgtCoDFish:
This will prevent us needing to add a new config file to be able to test a new version of istio

I deleted most of the config files because they were all identical. I kept the two which weren't the same as the others, just for historical reference.


inteon:
/approve
/lgtm

Makes sense to just have one config file, it has not changed for any of the latest istio versions.

PR #441, opened 2024-11-20
Upgrade go version to 1.23.1 to clean image scans.

tomelliot16:
Mostly just house keeping but should be a insignificant change.
[CVE-2024-34158](https://access.redhat.com/security/cve/cve-2024-34158)
[CVE-2024-34156](https://access.redhat.com/security/cve/CVE-2024-34156)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/cert-manager/istio-csr
 
-go 1.23.0
+go 1.23.1
 
 // https://github.com/darccio/mergo/blob/cde9f0ea26cccb1168ee3900cf8ca457bb928c3c/README.md#100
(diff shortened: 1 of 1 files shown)


SgtCoDFish:
Hey, thanks for raising this!

Our go versioning stuff is a bit complicated but this isn't actually the go version used to build. That's actually defined in a different repo (https://github.com/cert-manager/makefile-modules/blob/2742d1c08f02af255ab11ed900ffe70b12d69535/modules/tools/00_mod.mk#L164) and pulled in here via automation.

As such, this will actually be built with a newer version and this PR will have no effect, so I'm going to close it.

Thank you for getting involved though 🚀

PR #458, opened 2024-12-11
Fix helm chart typos

wallrj:
While testing istio-csr recently, I noticed various references to `approver-policy` in the Helm chart values.
I've fixed those and also:
 * Added missing full stops at the end of comments (for consistency with the majority of the Helm values comments).
 * Turned some unformatted URLs into markdown links, to save having to manually make those changes when we publish the Helm chart documentation. Here for example: https://docs.venafi.cloud/vaas/k8s-components/c-istiocsr-helmvalues/#tolerations
 * A couple of other drive by typo fixes.

### DIFF
--- a/deploy/charts/istio-csr/README.md
+++ b/deploy/charts/istio-csr/README.md
@@ -128,7 +128,7 @@ Create a Prometheus ServiceMonitor resource.
 > default
 > ```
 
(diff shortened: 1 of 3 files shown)


SgtCoDFish (approved):
/lgtm
/approve

Seems like obvious improvements to me, thanks 👍

Issue #176, opened 2022-08-02
certificateDuration is not used for the Istio CSR generated certificate requests


hari-vamsikrishna:
istio-csr generated certificates for the workloads is not using the "certificateDuration" parameter from the values.yaml. 
The certificaterequests generated by the istio-csr are generated with "duration: 24h0m0s" instead of using the "certificateDuration" parameter from the values.yaml.

JoshVanL:
Hi @hari-vamsikrishna, you can change the maximum requested duration using the istio-csr flag `--max-client-certificate-duration=1h` however is istio workloads request for shorter than that, then istio-csr will request that duration.

hari-vamsikrishna:
Hi @JoshVanL,
  Thank you for the super quick response 👍 . 
  Yes, i see the flag set as specified from the helm chart values.yaml --max-client-certificate-duration=720h
  But, the scene is, Istio-csr (with AWS PCA integration) and istio is deployed in the eks cluster. I have deployed the httpbin from the https://raw.githubusercontent.com/istio/istio/1.11.7/samples/httpbin/httpbin.yaml. And, I see that there is no cert duration requested/configured from the workload. 
  I think these certificates are automatically requested by istio-csr for the istio mtls (certificate being used in memory by Envoy) when istio-injection is enabled to the namespace where this workload is deployed. 
  The problem is the certificate is renewed for every day, which is getting costly
  So, how can we configure the certificate duration in this case?
  ref: https://github.com/cert-manager/istio-csr/blob/v0.4.0/docs/getting_started.md

volkorny:
Hi @hari-vamsikrishna,
Have you been able to find a solution to the issue?

volkorny:
We solved limit of 24h by specifying [SECRET_TTL](https://istio.io/latest/docs/reference/commands/pilot-agent/#secret-ttl)

PR #469, opened 2024-12-20
Add app.server.certificateDuration

volkorny:
Add app.server.certificateDuration

In our service mesh, we aim to increase the certificate duration beyond 24 hours. I noticed a similar request in https://github.com/cert-manager/istio-csr/issues/176

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -64,6 +64,8 @@ type Options struct {
 	// this value, this value will be used instead.
 	MaximumClientCertificateDuration time.Duration
 
+	ClientCertificateDuration time.Duration
+
 	// Authenticators configures authenticators to use for incoming CSR requests.
 	Authenticators AuthenticatorOptions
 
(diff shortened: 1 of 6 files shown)


pkg/server/server.go
```diff
@@ -212,6 +214,14 @@ func (s *Server) CreateCertificate(ctx context.Context, icr *securityapi.IstioCe
 		duration = s.opts.MaximumClientCertificateDuration
 	}
 
+	// If custom client duration is specified, override with the value.
+
+	if s.opts.ClientCertificateDuration > 0 {
+		duration = s.opts.ClientCertificateDuration
+	}
+
+	log.V(2).Info("Setting certificate duration", "duration", duration)
```
erikgb, line 223: If logging this, this should only be logged when overriding the duration. Why was log level 2 chosen? I would expect log level 3 after scanning through the code in this project.

pkg/server/server.go
```diff
@@ -212,6 +214,14 @@ func (s *Server) CreateCertificate(ctx context.Context, icr *securityapi.IstioCe
 		duration = s.opts.MaximumClientCertificateDuration
 	}
 
+	// If custom client duration is specified, override with the value.
+
+	if s.opts.ClientCertificateDuration > 0 {
+		duration = s.opts.ClientCertificateDuration
+	}
```
erikgb, line 221: I get a bit confused by this code that may modify `duration` twice (max and override)., and wonder if a shared switch statement would make this clearer.

erikgb (commented):
@volkorny Could you please describe a little better what you are trying to do here? Is `ValidityDuration` not given in the request which results in a default of 24h being applied? Or is it set, but you want all certificates to have the same `ValidityDuration`? With your proposed changes it seems like the latter. It is important to understand this to get good naming. I would suggest either `DefaultClientCertificateDuration` or `FixedClientCertificateDuration` - depending on what you are try to achieve.

It is also a bit harder to introduce a new field when we already have `MaximumClientCertificateDuration`. It should be documented how they relate to each other - depending on the implemented logic.

volkorny:
We solved limit of 24h by specifying [SECRET_TTL](https://istio.io/latest/docs/reference/commands/pilot-agent/#secret-ttl)
This PR is not actual.

PR #482, opened 2025-01-16
Use an alpha version of cert-manager to ensure fixed code is used

SgtCoDFish:
See https://github.com/cert-manager/istio-csr/pull/479#issuecomment-2587228710 for justification of this.

This should fix the failing govulncheck job, and prevent dependabot from trying to "upgrade" the cert-manager version to a broken version (which it's doing because it sees that 1.16.x is newer than v1.16.beta.0 - it doesn't see that we haven't backported [a fix](https://github.com/cert-manager/cert-manager/pull/7464) from master into the release-1.16 branch)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ go 1.23.0
 replace github.com/imdario/mergo => github.com/imdario/mergo v0.3.16
 
 require (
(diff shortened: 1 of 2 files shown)


SgtCoDFish:
This will be done in #483

PR #509, opened 2025-03-05
Bump istio module dependency + fix jose vulnerability

SgtCoDFish:
The istio dependency had a breaking change, which required renaming a function `NewFixedWatcher` to a new function in a different module.

The upstream break happened in https://github.com/istio/istio/pull/54629

NewTestWatcher seems correct since it has the same description as the old fixed watcher:
https://pkg.go.dev/istio.io/istio@v0.0.0-20250304232559-3173740c4a5c/pkg/config/mesh/meshwatcher#NewTestWatcher

This bump also simplifies so we depend on one version of the go-jose lib, and ensures that that jose lib is bumped to avoid CVE-2025-27144


inteon (commented):
/approve
/lgtm

Issue #534, opened 2025-04-23
helm value app.certmanager.additionalAnnotations interprets "JSON like" string as json and cannot render mainfests


jrhoward:
Hi I am testing Istio CSR integration with Venafi cloud and it works 🥇 . 

However we have a requirement to integrate specific annotation to Venafi CSR requests through the helm value `app.certmanager.additionalAnnotations`.

literally they should look like this in. the CRD Certificate object:

```yaml
annotations:
     venafi.cert-manager.io/custom-fields: |- 
                    [ {"name":"xx", "value": "xx" }, 
                     {"name":"xxx", "value": "xxx" }]
```

Creating a certificate with the following annotations works when we manually apply above example.

The helm chart however interprets it as a map and will not render charts. I have tried multiple way to force helm to not convert it to an array of type map and it always breaks, including:
- setting the value on one line, 
- using `--set-string` 
- escaping characters eg `\"`

but to no avail. 

using a helm values file:

```yaml

app:
  runtimeConfiguration:
    name: istio-issuer
  certmanager:
    additionalAnnotations:
        - name: "venafi.cert-manager.io/custom-fields"
          value: '[ { "Name": "xxxx", "Value": "xxxx" },{ "Name": "xxx", "Value": "xxx" } ]'
  tls:
    rootCAFile: "/var/run/secrets/istio-csr/ca.pem"
volumeMounts:
- name: root-ca
  mountPath: /var/run/secrets/istio-csr
volumes:
- name: root-ca
  secret:
    secretName: istio-root-ca
```

from the cli.

```sh
helm template cert-manager-istio-csr jetstack/cert-manager-istio-csr \
  --namespace cert-manager \
  --set "app.tls.rootCAFile=/var/run/secrets/istio-csr/ca.pem" \
  --set "volumeMounts[0].name=root-ca" \
  --set "volumeMounts[0].mountPath=/var/run/secrets/istio-csr" \
  --set "volumes[0].name=root-ca" \
  --set "volumes[0].secret.secretName=istio-root-ca" \
  --set "app.runtimeConfiguration.name=istio-issuer" \
  --set "app.certmanager.additionalAnnotations[0].name=venafi.cert-manager.io/custom-fields" \
  --set-string 'app.certmanager.additionalAnnotations[0].value=[{ "Name":"xxx", "Value":"xxx"},{ "Name":"xxxx", "Value":"xxxx"}]' \
  --debug
```
returns:

```sh
[...]

PR #537, opened 2025-04-24
fix: make handling of additionalAnnotations more robust:Jason Howard …

jrhoward:
…<jayhow@gmail.com>

The helm chart  interprets more complex additionalAnnotation incorrectly and will not render charts.

See linked issue for more detail:

https://github.com/cert-manager/istio-csr/issues/534

In summary this PR ensures annotation values passed to the csr agent are safe and ensures bootstrap Certificate annotation values are interpreted as a string.

### DIFF
--- a/deploy/charts/istio-csr/README.md
+++ b/deploy/charts/istio-csr/README.md
@@ -228,6 +228,9 @@ Takes key/value pairs in the format:
 additionalAnnotations:
   - name: custom.cert-manager.io/policy-name
     value: istio-csr
+  - name: venafi.cert-manager.io/custom-fields
+    value: |
+      [{ "Name": "field1", "Value": "value1" },{ "Name": "field2", "Value": "value2" }]
 ```
 #### **app.certmanager.issuer.enabled** ~ `bool`
 > Default value:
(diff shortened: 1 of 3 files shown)


jrhoward:
would some one be able to look at this please? it's a simple change to the helm charts @Alkemic @james-w @marcingy @munnerz

sitaramkm:
@jrhoward 
The better model to support your use-case would be configure `istio-csr` with a CA issuer so certs for mesh workloads can be issued locally. 
Injecting custom-field annotations into `CertificateRequest` implies that you are looking at having Venafi issue all certificates for mesh workloads. Managing & maintaining an inventory for large number of workload certs is not of much value. As you probably already know, we do not create a `Certificate` or a `Secret` resource when istio-csr signs incoming mesh workloads. In fact, our recommendation is to turn off `preserveCertificateRequests` for the mesh as it's typically cert with a short duration and are very often renewed. 
 
I agree that the PR itself is simple but the supporting usecase is not something I would recommend (that is assuming I am understanding the underlying usecase correctly) 
My recommendation if you plan to use open-source components

Create a CA issuer with Venafi as your issuer. As you will create this with a `Certificate` resource you can have all the annotations that Venafi expects. 
Configure this CA issuer in istio-csr. All mesh workloads will anchor back to the above CA issuer and it's root. 
You can also use policy-approver and create a few `CertificateRequestPolicy` resources if you want to constrain them to specific issuer/namespace, etc.

jrhoward:
Thanks @sitaramkm , have taken your advice on board and will be looking to change our strategy, would still be good to get this merged as it makes the passing of cli arguments to the CSR agent safer.

PR #586, opened 2025-08-12
Fix new noctx linter violations

erikgb:
This PR is to ensure that https://github.com/cert-manager/istio-csr/pull/584 will succeed in CI. The golangci-lint upgrade requires some code to be made more context-aware.

Framework updates are based on equivalent code copied from https://github.com/cert-manager/csi-driver-spiffe.


SgtCoDFish (approved):
/lgtm
/approve

Thank you!

PR #598, opened 2025-08-25
Upgrade Istio to release 1.27.0

erikgb:
Mainly to unblock https://github.com/cert-manager/istio-csr/pull/596, but generally a good idea to keep up with upstream releases.


hjoshi123:
Looks good to me.. /approve

maelvls:
Hey. Not very acquainted with Istio, but it looks like a simple enough change to approve. Seems like this code path is on the code path of most end-to-end tests, so I feel confident that nothing will break.

/lgtm
/approve

PR #728, opened 2026-01-31
Deprioritize resync operations

alandiegosantos:
In this controller, the re-sync of the istio-ca-root configmap is not a priority compared with the events derived from namespace operations or the change in the root CA. The resync might take tenths of minutes in clusters with thousands of namespaces, slowing the provisioning of the configmap on new namespaces.

So, this MR updates the controller to use WithLowPriorityWhenUnchanged, reducing the priority of the re-sync if the configmap is unchanged. PriorityQueues are default in controller-runtime v0.23, so no other change is necessary.

### DIFF
--- a/pkg/controller/configmap.go
+++ b/pkg/controller/configmap.go
@@ -108,11 +108,17 @@ func AddConfigMapController(ctx context.Context, log logr.Logger, opts Options) 
 	}
 
 	return ctrl.NewControllerManagedBy(opts.Manager).
-		// Reconcile ConfigMaps but only cache metadata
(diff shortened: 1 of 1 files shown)


ciaccotaco:
I tried to test this locally and got the following error. It is probably the same reason the tests are all failing. 

> `Error: failed to add CA root controller: one of For() or Named() must be called`

Coincidentally, I recently implemented changes in my environment to solve the same problem and created [PR 752](https://github.com/cert-manager/istio-csr/pull/752) with how I approached it.

This pull request and my pull request do two different things.  I think if this pull request can be updated to eliminate the error, then both can live side by side.

- PR 728 (this PR) improves queue fairness and startup scalability.
- [PR 752](https://github.com/cert-manager/istio-csr/pull/752) (Cache Root CA Bundle in ConfigMap Controller) caches the Root CA bundle to reduce per-reconcile overhead.

PR #769, opened 2026-03-30
Fix HasIssuerConfig to use RLock instead of Lock

archy-rock3t-cloud:
HasIssuerConfig only reads activeIssuerRef but acquires a write lock, unnecessarily blocking concurrent Sign calls.

### DIFF
--- a/pkg/certmanager/certmanager.go
+++ b/pkg/certmanager/certmanager.go
@@ -546,8 +546,8 @@ func (m *manager) WaitForIssuerConfig(ctx context.Context) {
 }
 
 func (m *manager) HasIssuerConfig() bool {
-	m.activeIssuerRefMutex.Lock()
-	defer m.activeIssuerRefMutex.Unlock()
+	m.activeIssuerRefMutex.RLock()
+	defer m.activeIssuerRefMutex.RUnlock()
 
 	return m.activeIssuerRef != nil
 }


archy-rock3t-cloud:
Bump — small concurrency fix (RLock for read-only path). Companion #768 adds tls.Provider unit tests. Could a maintainer take a look at both when there's time? Thanks!

skizot722:
Bump again - we're actually running up against this on our production clusters and would greatly appreciate the fix.  Thanks!

jq-cruise:
Bump again as the issue this PR is trying to fix has some production impact for us

We run `cert-manager-istio-csr` (v0.16.0) with a Vault-backed issuer. When Vault cert-signing latency was elevated (>10s under load), `kubelet` readiness probes on the `istio-csr` pod timed out. After three consecutive failures, the pod was marked unhealthy and restarted — even though the process was still functioning and actively signing certificates.

Readiness checks for `tls_provider` and `grpc_server` are wrapped with `deferCheckUntilIssuerConfig`, which calls `HasIssuerConfig()` on every probe ([cmd/app/app.go#L130-L142](https://github.com/cert-manager/istio-csr/blob/main/cmd/app/app.go#L130-L142)).

Today, `HasIssuerConfig()` takes an exclusive write lock while only reading state ([pkg/certmanager/certmanager.go#L548-L552](https://github.com/cert-manager/istio-csr/blob/main/pkg/certmanager/certmanager.go#L548-L552)). Meanwhile, `Sign()` holds a read lock for the entire duration of certificate issuance, including waiting on the issuer ([pkg/certmanager/certmanager.go#L193-L195](https://github.com/cert-manager/istio-csr/blob/main/pkg/certmanager/certmanager.go#L193-L195)).

Because the write lock waits for all readers to release, slow in-flight signing requests block readiness probes for the same duration. Switching `HasIssuerConfig()` to `RLock()` should allow probes to proceed concurrently with signing.

erikgb (commented):
This looks good, thanks @archy-rock3t-cloud! 🚀 

/lgtm
/approve

skizot722:
Thank you for the quick merge! Do you know when the next release containing this fix might be?

wallrj:
@archy-rock3t-cloud This change has been included in [v0.17.0](https://github.com/cert-manager/istio-csr/releases/tag/v0.17.0), which is now published.

Installation instructions are at https://cert-manager.io/docs/usage/istio-csr/installation/

If you are able to install the new release and verify the fix in your environment, that would be much appreciated. Thank you for the contribution.

*with claude fable-5*

skizot722:
Do you know when the v0.17.0 helm chart might be available?

erikgb:
> Do you know when the v0.17.0 helm chart might be available?

It is available when the release is built: https://quay.io/repository/jetstack/charts/cert-manager-istio-csr?tab=tags&tag=latest

Are you still using the legacy https repo?

skizot722:
> > Do you know when the v0.17.0 helm chart might be available?
> 
> It is available when the release is built: https://quay.io/repository/jetstack/charts/cert-manager-istio-csr?tab=tags&tag=latest
> 
> Are you still using the legacy https repo?

Ah ha! Yes, we were. Thank you!

Issue #786, opened 2026-04-21
Apply cluster API server TLS security profile to TLS servers (incl. curves / key exchange preferences)


arun717:
## Summary

istio-csr terminates TLS on one or more server endpoints (for example, the gRPC service and metrics). Today those settings may not follow a cluster-defined TLS security policy, and may rely partly on Go defaults for aspects such as cipher suites and curve / key-exchange preferences.

This issue tracks aligning **TLS server** configuration with the **cluster-wide policy exposed for the Kubernetes API server** on platforms that define it (for example, OpenShift’s `APIServer` / `tlsSecurityProfile`), so deployments stay consistent with the customer’s chosen profile (including **custom** profiles derived from a base type).

This request is scoped to **TLS servers hosted by istio-csr** (not client TLS to external services), consistent with the principle that server TLS should be driven by the same class of global knob where possible.

## Motivation

- **Central policy:** Operators and cluster admins increasingly configure TLS once at the cluster level instead of duplicating cipher/TLS version/curve policy in every component.
- **PQC-readiness:** Broader platforms are moving toward policies that favor TLS 1.3 and explicit control over algorithms; components that ignore the configured profile can lag behind the cluster’s security posture.
- **Consistency:** Hard-coded or implicit defaults can diverge from the profile enforced elsewhere on the cluster.

## Proposed direction

1. On clusters where an API server TLS security profile (or equivalent) is available, **watch** that configuration and **map** it to concrete
   `crypto/tls` (or equivalent) settings for istio-csr’s TLS servers.
2. Apply **all relevant** profile fields supported by the runtime — at minimum TLS protocol versions and cipher suites, and **curve /
   key-exchange preferences** where the stack allows them to be expressed (acknowledging current Go/library limitations where necessary).
3. On clusters **without** that API, keep **today’s** effective defaults unless/until a portable upstream mechanism exists (no surprise behavior change for generic Kubernetes).

## Scope

- In scope: TLS **servers** inside istio-csr, configuration wiring (flags / Helm / code), reconciliation when cluster policy changes, tests.
- Out of scope (unless maintainers want to split): TLS **client** behavior toward external CAs or workloads; unrelated refactors.

## Non-goals

- Replacing certificate issuance or trust bootstrap logic.
- Changing behavior on distributions that do not expose the configuration source used for this feature (unless explicitly agreed with maintainers).

## Acceptance criteria

[...]

PR #787, opened 2026-04-21
Support configurable min TLS version, cipher suites, and curves for istio-csr

arun717:
## Summary

Adds configurable TLS parameters for the istio-csr gRPC serving listener: minimum TLS version, cipher suites, and curve / key-exchange preferences. Uses Kubernetes-style TLS version names and k8s.io/component-base/cli/flag for min version and cipher suite parsing. Curve preferences use a local ParseCurvePreferences helper that accepts friendly names (X25519, P-256, etc.) and decimal tls.CurveID values. When unset, behavior matches previous defaults (TLS 1.2 minimum, Go default ciphers and curves).

Refs: https://github.com/cert-manager/istio-csr/issues/786

## What changed

- pkg/tls/tls.go — Apply optional MinVersion, CipherSuites, and CurvePreferences on both the outer tls.Config from Config() and the inner per-certificate config from fetchCertificate(). Enforce a TLS 1.2+ floor for MinVersion.
- pkg/tls/curves.go (+ pkg/tls/curves_test.go) — ParseCurvePreferences maps curve names and decimal IDs to []tls.CurveID.
- cmd/app/options/options.go — New flags: --serving-tls-min-version, --serving-tls-cipher-suites, --serving-tls-curve-preferences.
- deploy/charts/istio-csr/ — Optional Helm values wired through in values.yaml and templates/deployment.yaml; regenerated README.md and values.schema.json. Cipher suite docs note they only affect TLS 1.0–1.2 (TLS 1.3 cipher suites are not configurable in Go).
- pkg/tls/tls_test.go — Unit tests for invalid TLS options, provider creation, and serving TLS settings applied to returned tls.Config (inner and outer configs).

## Testing

- [ ]  make test-unit
- [ ] Manual: set Helm values or flags to a known policy, restart pods, verify with openssl s_client that version / ciphers / curves match expectations

## Notes / limitations

- Cipher suite configuration only affects TLS 1.0–1.2 connections; TLS 1.3 cipher suites are fixed by Go.
- Metrics and readiness remain plain HTTP on controller-runtime bind addresses; only the gRPC serving path uses these TLS settings.
- Settings take effect on pod restart (flag/Helm-driven), not via dynamic reconciliation.
- Curve parsing is custom rather than component-base’s TLSCurvePreferences (not available in v0.35.4; v0.36+ accepts numeric IDs only).

```release-note
istio-csr: the gRPC serving listener supports configurable minimum TLS version, cipher suites, and curve preferences (Helm values and CLI flags), defaulting to TLS 1.2 with Go defaults when unset.

### CODE REVIEW

File: cmd/app/options/options.go [hunk 1]
```diff
@@ -261,6 +261,24 @@ func (o *Options) addTLSFlags(fs *pflag.FlagSet) {
...
 			"Currently only RSA and ECDSA are supported. By default RSA is used.")
+
+	tlsCipherPossibleValues := cliflag.TLSCipherPossibleValues()
+	fs.StringSliceVar(&o.TLS.ServingTLSCipherSuites,
+		"serving-tls-cipher-suites", o.TLS.ServingTLSCipherSuites,
+		"Comma-separated list of cipher suites for the gRPC serving listener. "+
+			"If omitted, the default Go cipher suites are used. "+
+			"Possible values include: "+strings.Join(tlsCipherPossibleValues, ","))
+	tlsPossibleVersions := cliflag.TLSPossibleVersions()
+	fs.StringVar(&o.TLS.ServingTLSMinVersion,
```
bharath-b-rh, line 273: Should we also provide the option to configure the maximum TLS version?

pkg/tls/curves.go
```diff
@@ -0,0 +1,86 @@
...
+		id, err := parseOneCurveID(name)
+		if err != nil {
+			return nil, err
+		}
+		out = append(out, id)
+	}
+	return out, nil
+}
+
+func parseOneCurveID(name string) (tls.CurveID, error) {
```
bharath-b-rh: nit:
```suggestion
func parseCurveID(name string) (tls.CurveID, error) {
```

pkg/tls/curves_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"crypto/tls"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestParseCurvePreferences(t *testing.T) {
+	got, err := ParseCurvePreferences([]string{"X25519", "CurveP256"})
```
bharath-b-rh: nit: I feel it would make readability better, if written in table-driven tests format.

pkg/tls/provider_test.go
```diff
@@ -0,0 +1,68 @@
...
+	return nil
+}
+
+func (stubIssuerNotifier) HasIssuerConfig() bool { return true }
+
+func (stubIssuerNotifier) InitialIssuer() *cmmeta.IssuerReference { return nil }
+
+func TestNewProvider_InvalidServingTLS(t *testing.T) {
+	t.Parallel()
+	_, err := NewProvider(logr.Discard(), cmfake.New(), Options{
```
bharath-b-rh: nit: I feel it would make readability better, if written in table-driven tests format here and the next test case too.

pkg/tls/tls.go
```diff
@@ -293,11 +336,21 @@ func (p *Provider) Config(ctx context.Context) (*tls.Config, error) {
...
 
 		if conf != nil {
-			return &tls.Config{
-				MinVersion:         tls.VersionTLS12,
+			// Keep MinVersion enforcement local so static analyzers can prove we never
+			// build configs below TLS 1.2.
+			effectiveMinVersion := max(p.servingMinVersion, uint16(tls.VersionTLS12))
+			cfg := &tls.Config{
+				MinVersion:         effectiveMinVersion, // #nosec G402 -- value is clamped to TLS 1.2+ above.
 				GetConfigForClient: p.getConfigForClient,
```
bharath-b-rh, line 337: Should we initialize `GetConfigForClient` later, that is after setting `CipherSuites` and `CurvePreferences`, so that the client config also uses it?

cmd/app/options/options.go
```diff
@@ -261,6 +261,24 @@ func (o *Options) addTLSFlags(fs *pflag.FlagSet) {
 		"serving-signature-algorithm", "RSA",
 		"The type of signature algorithm to use when generating private keys. "+
 			"Currently only RSA and ECDSA are supported. By default RSA is used.")
+
+	tlsCipherPossibleValues := cliflag.TLSCipherPossibleValues()
+	fs.StringSliceVar(&o.TLS.ServingTLSCipherSuites,
+		"serving-tls-cipher-suites", o.TLS.ServingTLSCipherSuites,
+		"Comma-separated list of cipher suites for the gRPC serving listener. "+
+			"If omitted, the default Go cipher suites are used. "+
+			"Possible values include: "+strings.Join(tlsCipherPossibleValues, ","))
```
SgtCoDFish: suggestion: Worth calling out here that this doesn't apply for TLS 1.3. I think someone who's configuring TLS 1.3 only probably knows that, but I think it's good to be explicit in the flag docs.

deploy/charts/istio-csr/values.yaml
```diff
@@ -198,6 +198,19 @@ app:
     # Based on [NIST 800-204A recommendations (SM-DR13)](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-204A.pdf).
     certificateDuration: 1h
 
+    # Minimum TLS version for the gRPC serving listener (Kubernetes-style name, e.g. VersionTLS12).
+    # If unset, TLS 1.2 is used.
+    # +docs:property
+    # servingTLSMinVersion: VersionTLS12
+
+    # Cipher suites for the gRPC serving listener (TLS_CIPHER_SUITE names). If empty, Go defaults apply.
```
SgtCoDFish, line 206: suggestion: (as before) Worth calling out here that this doesn't apply for TLS 1.3. I think someone who's configuring TLS 1.3 only probably knows that, but I think it's good to be explicit in the value docs.

pkg/tls/tls.go
```diff
@@ -116,6 +130,12 @@ type Provider struct {
 
 	cm certmanager.Signer
 
+	servingMinVersion        uint16
+	servingCipherSuites      []uint16
+	servingCurvePreferences  []tls.CurveID
+	servingApplyCipherSuites bool
+	servingApplyCurvePrefs   bool
```
SgtCoDFish: suggestion: I'd suggest we remove these variables and just check `len(cipherSuites) > 0` and `len(curves) > 0` whenever we need to.

I think having separate variables for the same state could be confusing and error prone.

pkg/tls/provider_test.go
```diff
@@ -0,0 +1,120 @@
...
+					return
+				}
+				require.Error(t, err)
+				return
+			}
+
+			require.NoError(t, err)
+			require.NotNil(t, p)
+			require.Equal(t, tt.applyCipherSuites, p.servingApplyCipherSuites)
+			require.Equal(t, tt.applyCurvePreferences, p.servingApplyCurvePrefs)
```
SgtCoDFish: suggestion: I think this test is a bit brittle. It doesn't verify the actual TLS config that gets applied - it just checks values on the struct - so if we forget to set the cipher suites on the TLS config in future code, we just won't notice.

Maybe a good way to test without breaking stuff would be to add a case that seeds p.tlsConfig (or hits fetchCertificate via the fake signer) and asserts cfg.MinVersion, cfg.CipherSuites, cfg.CurvePreferences on the returned tls.Config.

pkg/tls/provider_test.go
SgtCoDFish, line 1: suggestion(non-blocking): I think the contents of this file could reasonably be in `tls_test.go`, to save adding a new file.

pkg/tls/curves.go
SgtCoDFish, line 1: question: Could we use `k8s.io/component-base/cli/flag` to help with some of this parsing stuff? Specifically https://pkg.go.dev/k8s.io/component-base/cli/flag#TLSCipherSuites ?

SgtCoDFish (changes requested):
Got a few things which i think would help here. Might need another review loop after!

Thanks for raising this!

pkg/tls/tls_test.go
```diff
@@ -0,0 +1,252 @@
...
+				return
+			}
+
+			require.NoError(t, err)
+			require.NotNil(t, p)
+		})
+	}
+}
+
+func TestProviderServingTLSConfig(t *testing.T) {
```
SgtCoDFish, line 184: suggestion: no tests here for an invalid / typo'd cipher suite, right? would be worth adding one!

deploy/charts/istio-csr/values.yaml
```diff
@@ -198,6 +198,20 @@ app:
...
+    # Only affects TLS 1.0–1.2; TLS 1.3 cipher suites are not configurable in Go.
+    # +docs:property
+    servingTLSCipherSuites: []
```
SgtCoDFish, line 212: suggestion: This is the value that people are going to set if they want to modify the cipher suites, but it doesn't give any indication of what values are acceptable. Could we add a link to some docs showing what can be passed?

cmd/app/options/options.go
```diff
@@ -261,6 +261,25 @@ func (o *Options) addTLSFlags(fs *pflag.FlagSet) {
 		"serving-signature-algorithm", "RSA",
 		"The type of signature algorithm to use when generating private keys. "+
 			"Currently only RSA and ECDSA are supported. By default RSA is used.")
+
+	tlsCipherPossibleValues := cliflag.TLSCipherPossibleValues()
+	fs.StringSliceVar(&o.TLS.ServingTLSCipherSuites,
+		"serving-tls-cipher-suites", o.TLS.ServingTLSCipherSuites,
+		"Comma-separated list of cipher suites for the gRPC serving listener. "+
+			"If omitted, the default Go cipher suites are used. "+
```
SgtCoDFish, line 268: suggestion(AI): Reviewed the below comment and agree with it. Same caveats could apply to the helm value documentation.

Cipher suite doc doesn't warn about insecure options

cliflag.TLSCipherPossibleValues() includes suites like TLS_RSA_WITH_RC4_128_SHA (InsecureTLSCiphers()). The flag help just enumerates all of them. A user copying from --help output could unknowingly enable a broken cipher. Consider filtering to [PreferredTLSCipherNames()](https://pkg.go.dev/k8s.io/component-base/cli/flag#PreferredTLSCipherNames) in the help text, or noting "some listed values are considered insecure — see k8s.io/component-base/cli/flag."

SgtCoDFish (changes requested):
Few more comments - definitely getting closer on this!

pkg/tls/tls.go
```diff
@@ -424,8 +474,11 @@ func (p *Provider) fetchCertificate(ctx context.Context) (time.Time, error) {
 	// this provider. This config will serve using the just signed certificate
 	// and private key. Mutually authenticate incoming client requests based if a
 	// certificate is present.
-	p.tlsConfig = &tls.Config{
-		MinVersion:   tls.VersionTLS12,
+	// Keep MinVersion enforcement local so static analyzers can prove we never
+	// build configs below TLS 1.2.
+	effectiveMinVersion := max(p.servingMinVersion, uint16(tls.VersionTLS12))
+	inner := &tls.Config{
+		MinVersion:   effectiveMinVersion, // #nosec G402 -- value is clamped to TLS 1.2+ above.
```
SgtCoDFish: suggestion(AI): I chewed on this suggestion from claude for a while and I think I agree with it. Having two separate clamps is likely to cause issues in the future. Admittedly, unlikely to actually ever change, but if we're changing something else it seems worth this refactor?

---

Duplicated TLS-config-building logic (pkg/tls/tls.go)

The "clamp to TLS 1.2, then conditionally apply cipher suites/curve preferences" block is duplicated verbatim in two places:
- Config() (lines 338–349)
- fetchCertificate() (lines 479–504)

Including the identical `#nosec G402` comment/pragma in both spots. If a future change updates one (e.g. adding another field, or changing the floor logic) it's easy to forget the other, and gosec suppression comments duplicated like this are exactly the kind of thing that silently goes stale. Consider extracting a small helper, e.g.:

```go
func (p *Provider) applyTLSSecuritySettings(cfg *tls.Config) {
    cfg.MinVersion = max(p.servingMinVersion, uint16(tls.VersionTLS12)) // #nosec G402 -- clamped above
    if len(p.servingCipherSuites) > 0 {
        cfg.CipherSuites = p.servingCipherSuites
    }
    if len(p.servingCurvePreferences) > 0 {
        cfg.CurvePreferences = p.servingCurvePreferences
    }
}
```

and call it from both sites.

deploy/charts/istio-csr/values.yaml
```diff
@@ -198,6 +198,23 @@ app:
     # Based on [NIST 800-204A recommendations (SM-DR13)](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-204A.pdf).
     certificateDuration: 1h
 
+    # Minimum TLS version for the gRPC serving listener (Kubernetes-style name, e.g. VersionTLS12).
+    # If unset, TLS 1.2 is used.
```
SgtCoDFish: suggestion: I think it'd be good to leave ourselves the freedom to bump the default in the future. We could reserve the right here.


```suggestion
    # Minimum TLS version for the gRPC serving listener (Kubernetes-style name, e.g. VersionTLS12).
    # If unset, TLS 1.2 is used but a future version will increase the default.
```

deploy/charts/istio-csr/values.yaml
```diff
@@ -198,6 +198,23 @@ app:
...
+    # +docs:property
+    # servingTLSMinVersion: VersionTLS12
+
+    # Cipher suites for the gRPC serving listener (TLS_CIPHER_SUITE names). If empty, Go defaults apply.
+    # Only affects TLS 1.0–1.2; TLS 1.3 cipher suites are not configurable in Go.
```
SgtCoDFish: suggestion: Remove reference to TLS 1.0-1.2 - we should only support TLS 1.2 and TLS 1.3

SgtCoDFish (changes requested):
Sorry for delay - few more bits. I recognise this is a bit nitpicky but I think it's right to do.

cmd/app/options/options.go
```diff
@@ -261,6 +261,25 @@ func (o *Options) addTLSFlags(fs *pflag.FlagSet) {
...
+			"Only affects TLS 1.2; TLS 1.3 cipher suites are not configurable in Go. "+
+			"Preferred values: "+strings.Join(cliflag.PreferredTLSCipherNames(), ", ")+". "+
+			"Insecure values: "+strings.Join(cliflag.InsecureTLSCipherNames(), ", ")+".")
+	tlsPossibleVersions := cliflag.TLSPossibleVersions()
+	fs.StringVar(&o.TLS.ServingTLSMinVersion,
+		"serving-tls-min-version", o.TLS.ServingTLSMinVersion,
+		"Minimum TLS version for the gRPC serving listener. "+
+			"If omitted, TLS 1.2 is used (matching the previous hard-coded default). "+
+			"Possible values: "+strings.Join(tlsPossibleVersions, ", "))
```
lunarwhite, line 277: This CLI flag help text is outdated. It should align with what's suggested in https://github.com/cert-manager/istio-csr/pull/787#discussion_r3713920317

pkg/tls/tls.go
```diff
@@ -101,6 +102,20 @@ type Options struct {
 	// ServingSignatureAlgorithm is the type of key of serving signature algorithm
 	// used, RSA or ECDSA, The default is RSA.
 	ServingSignatureAlgorithm string
+
+	// ServingTLSMinVersion is the minimum TLS version for the gRPC listener,
+	// using Kubernetes-style version names (for example VersionTLS12). Empty
+	// selects TLS 1.2, matching the historical default.
```
lunarwhite: Ditto.

pkg/tls/tls.go
```diff
@@ -309,6 +350,20 @@ func (p *Provider) Config(ctx context.Context) (*tls.Config, error) {
 	}
 }
 
+// applyTLSSecuritySettings sets MinVersion (clamped to TLS 1.2+), CipherSuites,
+// and CurvePreferences from the provider's serving TLS options onto cfg.
+func (p *Provider) applyTLSSecuritySettings(cfg *tls.Config) {
+	// Keep MinVersion enforcement local so static analyzers can prove we never
+	// build configs below TLS 1.2.
+	cfg.MinVersion = max(p.servingMinVersion, uint16(tls.VersionTLS12)) // #nosec G402 -- value is clamped to TLS 1.2+.
```
lunarwhite:
```suggestion
	cfg.MinVersion = max(p.servingMinVersion, uint16(tls.VersionTLS12))
```
I feel this annotation is not necessary. You could double confirm by deleting it and re-running with actually-pinned golangci-lint v2.12.2.

pkg/tls/curves.go
lunarwhite: Same question here. I think we should actually wire this into `TLSCurvePreferences` rather than parsing decimal IDs ourselves. `TLSCurvePreferences` already validates range, rejects duplicate entries, and rejects IDs Go's crypto/tls doesn't recognize (via a `curve.String()` prefix check): https://github.com/kubernetes/component-base/blob/v0.36.3/cli/flag/curves_flag.go#L29 None of which the PR's decimal-ID branch does.

Rather than re-deriving that validation here, could we keep just the friendly-name mapping (name -> ID alias table) and delegate the rest?

lunarwhite:
In PR description you mentioned

> Manual: set Helm values or flags to a known policy, restart pods, verify with openssl s_client that version / ciphers / curves match expectations

which is good. It'd be better if you also paste the actual test steps you performed and corresponding results, to show how would this feature work end-to-end. That will gain us great confidence to merge :)

arun717:
Thanks @lunarwhite
Please find below the manual e2e verification I ran against this PR (image built from this branch) on a local cluster.

Setup

1. Installed cert-manager (jetstack/cert-manager, CRDs enabled).

2. Created the usual bootstrap Issuer/CA in istio-system (selfsigned → istio-ca Certificate/Issuer).

3. Built cert-manager-istio-csr from this PR and deployed the local Helm chart with:

```
 certmanager:
    namespace: istio-system
    issuer:
      group: cert-manager.io
      kind: Issuer
      name: istio-ca
  tls:
    istiodCertificateEnable: false
    servingTLSMinVersion: VersionTLS12
    servingTLSCipherSuites:
      - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
      - TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
    servingTLSCurvePreferences:
      - X25519
      - CurveP256
```

4. Confirmed the pod args included:

- --serving-tls-min-version=VersionTLS12
- --serving-tls-cipher-suites=TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
- --serving-tls-cipher-suites=TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
- --serving-tls-curve-preferences=X25519
- --serving-tls-curve-preferences=CurveP256       

5. Port-forwarded the gRPC Service and probed with openssl s_client: 
`oc port-forward -n cert-manager svc/cert-manager-istio-csr 15443:443`

(OpenSSL 3 disables TLS ≤1.1 by default, so for those negative cases I used a temporary OPENSSL_CONF with MinProtocol = TLSv1 / CipherString = DEFAULT:@SECLEVEL=0.)

Results

| Probe | Expectation | Result |
|-------|-------------|--------|
| `-tls1` / `-tls1_1` | Reject below minimum TLS version | ❌ **FAIL** — `tlsv1 alert protocol version` (alert 70) |
| `-tls1_2` | Succeed | ✅ **OK** — TLS 1.2, `ECDHE-RSA-AES128-GCM-SHA256` |
| `-tls1_3` | Succeed (minimum TLS 1.2 still allows TLS 1.3) | ✅ **OK** — TLS 1.3, `TLS_AES_128_GCM_SHA256` |
| TLS 1.2 + `ECDHE-RSA-AES128-GCM-SHA256` | Allowed cipher | ✅ **OK** |
| TLS 1.2 + `ECDHE-RSA-AES256-GCM-SHA384` | Allowed cipher | ✅ **OK** |
| TLS 1.2 + `ECDHE-RSA-AES128-SHA` | Cipher not in configured list | ❌ **FAIL** — Handshake failure (alert 40) |
| TLS 1.2 + `-curves X25519` | Allowed curve | ✅ **OK** — `named_curve: ecdh_x25519`, Peer Temp Key: `X25519` |
| TLS 1.2 + `-curves P-256` | Allowed curve | ✅ **OK** |
| TLS 1.2 + `-curves P-384` | Curve not in configured preferences | ❌ **FAIL** — Handshake failure (alert 40) |

This matches the intended behavior
Helm/CLI TLS policy is applied on the gRPC serving listener, TLS < 1.2 is rejected, configured TLS 1.2 ciphers/curves are enforced, and TLS 1.3 still works when min version is 1.2 (Go-managed TLS 1.3 ciphers, as documented).

SgtCoDFish (approved):
/lgtm
/approve

I feel like this is the kind of thing that could get reviewed over and over but ultimately I don't think anyone gains from that. There are probably still things I'd change here but I just don't think it matters too much - as such, I think my default is to approve because this improves things. Thank you for implementing this (and for your patience in getting this over the line!)

PR #817, opened 2026-05-28
Fix pipefail issue

inteon:
Fixes a pipefail failure in the CA rotation test caused by openssl x509 closing stdin after reading the first cert in a chain, which sent SIGPIPE to the upstream base64 -d. The fix uses process substitution so openssl reads from a file-like input instead of a pipe.

It is unclear what caused the `certificateChain` to be multi-cert now.

Generated by Claude


erikgb (commented):
/lgtm
/approve

Thanks Tim (and Claude)! 🚀

PR #834, opened 2026-06-16
feat(chart): expose priorityClassName in the istio-csr chart

yugstar:
## What

Exposes a `priorityClassName` value in the istio-csr Helm chart and wires it into the Deployment.

## Why

Every other cert-manager chart already lets operators set a `priorityClassName` on the workload (cert-manager, trust-manager, approver-policy, csi-driver, csi-driver-spiffe, google-cas-issuer), using the same value, comment and template idiom. istio-csr was the only one missing it, so operators who rely on PriorityClasses to protect this control-plane component from preemption had to fork the chart.

This adds the value with the exact same wording and `{{- with .Values.priorityClassName }}` idiom as the sibling charts.

## Behaviour

Default is `""`, so the rendered Deployment is unchanged. Verified:
- `helm template` default → no `priorityClassName` rendered.
- `helm template --set priorityClassName=system-cluster-critical` → renders correctly.
- `values.schema.json` and `README.md` regenerated via `make generate-helm-schema generate-helm-docs`.
- `make verify-helm-values`, `verify-helm-lint`, `verify-helm-unittest` pass.

```release-note
The istio-csr Helm chart now supports setting `priorityClassName` on the deployment.
```

### DIFF
--- a/deploy/charts/istio-csr/README.md
+++ b/deploy/charts/istio-csr/README.md
@@ -542,6 +542,13 @@ resources:
     cpu: 100m
     memory: 128Mi
 ```
+#### **priorityClassName** ~ `string`
+> Default value:
+> ```yaml
+> ""
+> ```
+
+Configure the priority class of the pod. For more information, see [PriorityClass](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#priorityclass).
 #### **securityContext** ~ `object`
 > Default value:
 > ```yaml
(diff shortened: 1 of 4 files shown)


erikgb (commented):
Thanks @yugstar! 🚀

/lgtm
/approve

wallrj:
@yugstar This change has been included in [v0.17.0](https://github.com/cert-manager/istio-csr/releases/tag/v0.17.0), which is now published.

Installation instructions are at https://cert-manager.io/docs/usage/istio-csr/installation/

If you are able to install the new release and verify the change in your environment, that would be much appreciated. Thank you for the contribution.

*with claude fable-5*

